PR #19, opened 2017-10-06
Allow spaces in step names

ruksi:


akx:
Okay, so this is actually a matter of shell usage, so I'm closing this.

For non-ambiguous step names with spaces, one needs to quote the step name:

```
$ vh exec run "My Awesome Step With Multiple Words" --...
```

PR #25, opened 2018-03-26
More small improvements and groundwork

akx:
This lays some groundwork for future additions.


akx:
Seems Codecov is being wonky, I'll see what's up there...

PR #26, opened 2018-03-28
Add parcel command

akx:
This PR adds a brand-spankin'-new `parcel` command, which creates a self-contained copy of the project and its dependencies.


ruksi:
GTG after fixing the tests.

PR #33, opened 2018-07-11
Be more human-friendly about parameter/input names

### LINKED ISSUE #29: Step parameters with spaces don't work
Passing step parameters with spaces does't work, it does use such default parameters fine though.

Maybe we should require a separate `id` attribute in params for this? Or should we just slugify the name? Or enforce more strict naming for parameters?

Example below.

===========

```
(valohai-cli) bog:darknet-example ruksi$ vh exec run --step="Generate text" --'Textual Seed'=1
Usage: vh execution run [OPTIONS] [ARGS]...

Error: no such option: --Textual Seed
```

akx:
Fixes #29

* This turns parameter/input names like "Input Data" into "--input-data"
* It also adds Levenshtein distance based spell checking, so "--ipnutd-ada" will yield a suggestion for "--input-data".


valohai_cli/commands/execution/run.py
```diff
@@ -170,6 +171,14 @@ def resolve_commit(self, commit):
 
         return commit
 
+    def make_parser(self, ctx):
+        parser = super(RunCommand, self).make_parser(ctx)
+        # This is somewhat naughty, but allows us to easily hook into here.
+        # Besides, FriendlyOptionParser does inherit from OptionParser anyway,
+        # and just overrides that one piece of behavior...
+        parser.__class__ = FriendlyOptionParser
```
ruksi: 👌

valohai_cli/git.py
```diff
@@ -12,7 +12,7 @@ def check_git_output(args, directory):
             stderr=subprocess.STDOUT,
         )
     except subprocess.CalledProcessError as cpe:
-        if cpe.returncode == 128 and 'Not a git repository' in cpe.output.decode():
+        if cpe.returncode == 128 and 'not a git repository' in cpe.output.decode().lower():
```
ruksi: good catch

PR #44, opened 2018-11-23
Add --title to execution parameters

JuhaKiili:
Also add -i for shorter version of --image


JuhaKiili:
Dependency with: https://github.com/valohai/roi/pull/782

akx:
@JuhaKiili Is this WIP only due to the dep on the API PR?

JuhaKiili:
> @JuhaKiili Is this WIP only due to the dep on the API PR?

Yes it is good to go now

PR #45, opened 2018-11-26
Move ad-hoc packaging after all input validation

### LINKED ISSUE #43: Input/command validation should be done before ad-hoc packaging
It otherwise can take a long time for packaging to be done, only for the user to realize that they forgot a flag.

JuhaKiili:
Fixes #43

### DIFF
--- a/valohai_cli/commands/execution/run.py
+++ b/valohai_cli/commands/execution/run.py
@@ -217,13 +217,11 @@ def run(ctx, step, commit, environment, watch, title, adhoc, image, args):
     matched_step = match_step(config, step)
     step = config.steps[matched_step]
 
-    if adhoc:
-        commit = create_adhoc_commit(project)['identifier']
-
     rc = RunCommand(project, step, commit=commit, environment=environment, watch=watch, image=image, title=title)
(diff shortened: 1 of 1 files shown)


valohai_cli/commands/execution/run.py
```diff
@@ -217,12 +217,13 @@ def run(ctx, step, commit, environment, watch, title, adhoc, image, args):
     matched_step = match_step(config, step)
     step = config.steps[matched_step]
 
-    if adhoc:
-        commit = create_adhoc_commit(project)['identifier']
+    rc = RunCommand(project, step, commit=None, environment=environment, watch=watch, image=image, title=title)
```
akx: Couldn't we leave this line as-is? The way I see it, `commit` will be None if `adhoc` is true, so we only need to set it post-fact if we're doing `adhoc`.

That is,

```python
rc = RunCommand(..., commit=commit, ...)
with rc...:
   if adhoc:
       commit = ...
       rc.commit = commit
   return rc.invoke(...)
```

JuhaKiili: Pushed fix

akx: Much more elegant! Cheers. :)

PR #51, opened 2019-02-14
Move linting logic from valohai-cli to valohai-yaml for reusability

JuhaKiili:
Dependency on https://github.com/valohai/valohai-yaml/pull/13


valohai_cli/commands/lint.py
```diff
@@ -20,38 +18,16 @@ def validate_file(filename):
...
-        obj_path = [str(el) for el in error.path]
-        click.echo('  {validator} validation on {schema_path}: {message} ({path})'.format(
-            validator=click.style(error.validator.title(), bold=True),
-            schema_path=click.style('.'.join(simplified_schema_path), bold=True),
-            message=click.style(error.message, fg='red'),
-            path=click.style('.'.join(obj_path), bold=True),
-        ))
+    click.secho('%s: %d errors, %d warnings' % (filename, lr.error_count, lr.warning_count), fg='yellow', bold=True)
+    for message in lr.messages:
+        click.echo('  {message}'.format(message=message['message']))
```
akx: This should probably have a prefix to distinguish between warnings and errors. Forgot about that in the MVP implementation.

Something as simple as 

```python
click.echo('  {type}: {message}'.format(**message))
```
will probably do.

JuhaKiili: Fix pushed

requirements-dev.txt
```diff
@@ -2,3 +2,4 @@ pytest
 pytest-cov
 requests-mock
 tox
+valohai-yaml==0.8.0
```
akx: `valohai-yaml` is a hard requirement, not a development requirement, so please just bump the specifier line in https://github.com/valohai/valohai-cli/blob/90e8d5e1cc2d5df6bcc08c7920f06de5a04eae60/setup.py#L16 to `valohai-yaml>=0.8.0` :)

tox.ini
```diff
@@ -8,3 +8,4 @@ deps =
     pytest
     pytest-cov
     requests-mock
+    valohai-yaml==0.8.0
```
akx: Same here :)

PR #52, opened 2019-02-26
Support multiple files per input (fixes #24)

### LINKED ISSUE #24: Support multiple files per input
Support for this seems to be lacking in the CLI, even if the web UI and API support it.

JuhaKiili:


valohai_cli/commands/execution/run.py
```diff
@@ -71,6 +71,7 @@ def __init__(self, project, step, commit, environment, image, title, watch):
         super(RunCommand, self).__init__(
             name=sanitize_name(step.name.lower()),
             callback=self.execute,
+            epilog='Multiple files per input: --my-input: <myurl> --my-input: <myotherurl>',
```
akx: This seems misleading with those colons...

> `--my-input=url --my-input=url ...`

maybe?

tests/commands/execution/test_run.py
```diff
@@ -83,7 +83,7 @@ def test_run(runner, logged_in_and_linked, monkeypatch, pass_param, pass_input,
 
     if pass_input:
         args.append('--in1=http://url')
-        values['inputs'] = {'in1': 'http://url'}
+        values['inputs'] = {'in1': ['http://url']}
```
akx: Could we also get a test that actually passes in multiple inputs?

valohai_cli/commands/execution/run.py
```diff
@@ -107,8 +108,9 @@ def convert_input_to_option(self, input):
         option = click.Option(
             param_decls=list(generate_sanitized_options(input.name)),
             required=(input.default is None and not input.optional),
-            default=input.default,
+            default=[input.default],
```
akx: This should probably be `default=([input.default] if input.default else []),`; otherwise you'd get `default=[None]`...

valohai_cli/commands/execution/run.py
JuhaKiili: Fix pushed

tests/commands/execution/test_run.py
JuhaKiili: Fix pushed. I included it in the same test, though.

valohai_cli/commands/execution/run.py
JuhaKiili: Fix pushed

PR #54, opened 2019-03-05
A smorgasbord of improvements

akx:


valohai_cli/commands/execution/run.py
```diff
@@ -208,18 +227,19 @@ def make_parser(self, ctx):
 )
 @click.argument('step')
 @click.option('--commit', '-c', default=None, metavar='SHA', help='The commit to use. Defaults to the current HEAD.')
-@click.option('--environment', '-e', default=None, help='The environment UUID or slug to use.')
+@click.option('--environment', '-e', default=None, help='The environment UUID or slug to use (see `vh env`)')
 @click.option('--image', '-i', default=None, help='Override the Docker image specified in the step.')
 @click.option('--title', '-t', default=None, help='Title of the execution.')
 @click.option('--watch', '-w', is_flag=True, help='Start `exec watch`ing the execution after it starts.')
+@click.option('--env', '-E', 'environment_variables', multiple=True, help='Add environment variable (NAME=VALUE). May be repeated.')
```
ruksi: I would go with `--var` and `-v` instead of `--env` and `-E`

Mainly to reduce confusion with `--environment` and `-e`

akx: Yeah, I guess.

The reason for this nomenclature was to keep it in line with Docker's syntax, but... yeah, `--var` will do.

PR #57, opened 2019-03-12
Add ability to sync/download execution outputs live 

JuhaKiili:
Example commands:

Download and keep waiting for output files of latest execution
`vh outputs latest --sync --download=myfolder`

Start a new execution and keep waiting for output files
`vh execution run train --sync=myfolder`

### DIFF
--- a/valohai_cli/commands/execution/outputs.py
+++ b/valohai_cli/commands/execution/outputs.py
@@ -10,33 +10,65 @@ from valohai_cli.messages import success, warn
 from valohai_cli.table import print_table
 from valohai_cli.utils import force_text
 from valohai_cli.utils.cli_utils import counter_argument
-
+from valohai_cli.api import request
+from valohai_cli.consts import complete_execution_statuses
 
 @click.command()
 @counter_argument
 @click.option('--download', '-d', type=click.Path(file_okay=False),
-              help='download files to this directory (by default, don\'t download)', default=None)
-@click.option('--filter-download', '-f', help='download only files matching this glob', default=None)
-def outputs(counter, download, filter_download):
+              help='Download files to this directory (by default, don\'t download).', default=None)
+@click.option('--filter-download', '-f', help='Download only files matching this glob.', default=None)
+@click.option('--force', is_flag=True, help='Download all files even if they already exist.')
+@click.option('--sync', '-s', is_flag=True, help='Keep watching for new output files to download.')
+def outputs(counter, download, filter_download, force, sync):
     """
     List and download execution outputs.
     """
     execution = get_project(require=True).get_execution_from_counter(counter=counter)
-    outputs = execution.get('outputs', ())
-    if not outputs:
-        warn('The execution has no outputs.')
-        return
-    print_table(outputs, ('name', 'url', 'size'))
+
+    if sync:
+        watch(counter, force, filter_download, download)
+    else:
+        outputs = execution.get('outputs', ())
+        if not outputs:
+            warn('The execution has no outputs.')
+            return
+        print_table(outputs, ('name', 'url', 'size'))
+        if download:
+            outputs = filter_outputs(outputs, download, filter_download, force)
+            download_outputs(outputs, download, force, show_success_message=True)
+
+
+def watch(counter, force, filter_download, download):
     if download:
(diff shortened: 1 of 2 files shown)


valohai_cli/commands/execution/outputs.py
```diff
@@ -57,8 +76,9 @@ def download_outputs(outputs, output_path):
                     prog.update(len(chunk))
                     outf.write(chunk)
     duration = time.time() - start_time
-    success('Downloaded {n} outputs ({size} bytes) in {duration} seconds'.format(
-        n=len(outputs),
-        size=total_size,
-        duration=round(duration, 2),
-    ))
+    if(show_success_message):
```
akx: Extraneous parens.

valohai_cli/commands/execution/run.py
```diff
@@ -232,14 +241,16 @@ def make_parser(self, ctx):
 @click.option('--title', '-t', default=None, help='Title of the execution.')
 @click.option('--watch', '-w', is_flag=True, help='Start `exec watch`ing the execution after it starts.')
 @click.option('--var', '-v', 'environment_variables', multiple=True, help='Add environment variable (NAME=VALUE). May be repeated.')
+@click.option('--sync', '-s', type=click.Path(file_okay=False), help='Download execution outputs to DIRECTORY.', default=None)
```
akx: This is mutually exclusive with `--watch`, it looks like (or at least the behavior will be strange if both are set).

valohai_cli/commands/execution/outputs.py
```diff
@@ -15,28 +15,47 @@
...
-    if not outputs:
-        warn('Nothing to download.')
-        return
+        if not sync:
+            print_table(outputs, ('name', 'url', 'size'))
+
+        if download:
+            if filter_download:
+                outputs = [output for output in outputs if fnmatch(output['name'], filter_download)]
+            if not force or sync:
```
akx: Can you rearrange this to `if sync or not force:` ? (I at least assume that's the point here.)

valohai_cli/commands/execution/outputs.py
```diff
@@ -15,28 +15,47 @@
...
+                # Do not download files that already exist
+                outputs = [output for output in outputs if not os.path.isfile(os.path.join(download, output['name']))]
+            if outputs:
+                download_outputs(outputs, download, force, show_success_message=not sync)
+            elif not sync:
+                warn('No new files to download (use --force to redownload).')
```
akx: Will this appear on every 3-second cycle? Is that alright?

valohai_cli/commands/execution/outputs.py
```diff
@@ -15,28 +15,47 @@
...
-            outputs = [output for output in outputs if fnmatch(output['name'], filter_download)]
-        download_outputs(outputs, download)
+    if sync:
+        if download:
+            print("Downloading to: %s\nWaiting for new outputs..." % download)
+        else:
+            warn('Target folder is not set. Use --download to set it.')
+            return
 
+    while True:
```
akx: Sync mode and regular mode are somewhat strangely tangled here... it might be worth it to at least see if they could be separated without too much duplication.

valohai_cli/commands/execution/outputs.py
JuhaKiili: Fix pushed

valohai_cli/commands/execution/run.py
JuhaKiili: Pushed warning saying "not yet supported" as I think we should support it at some point.

valohai_cli/commands/execution/outputs.py
JuhaKiili: Pushed fix

valohai_cli/commands/execution/outputs.py
JuhaKiili: It didn't, because for not sync there was no cycle, but I refactored better split between sync and no-sync anyway...

valohai_cli/commands/execution/outputs.py
JuhaKiili: Pushed refactor

PR #61, opened 2019-03-20
Add more information to invalid environment error

JuhaKiili:


valohai_cli/exceptions.py
```diff
@@ -21,6 +21,9 @@ def __init__(self, response):
         if '<!DOCTYPE html>' in response.text:
             # Don't shower the user with a blob of HTML
             text = 'Internal error'
+        elif 'is not a valid UUID' in response.text:
```
akx: This error _could_ happen with other UUID-indexed things such as data stores... Also, commit 42bb828fe958a6c6f in the app itself will likely rephrase the error.

JuhaKiili: How would you recommend to differentiate this specific error from the rest?

akx:
Superseded by #76.

PR #85, opened 2019-06-11
Use environment variable default values from CLI (fixes #67)

### LINKED ISSUE #67: CLI does not honor default environment variables in step
They're probably currently only populated into the GUI.

JuhaKiili:


valohai_cli/commands/execution/run/dynamic_run_command.py
```diff
@@ -80,6 +80,9 @@ def __init__(self,
             self.params.append(self.convert_param_to_option(parameter))
         for input in step.inputs.values():
             self.params.append(self.convert_input_to_option(input))
+        for envvar in step.environment_variables.items():
```
akx: I'd like to see this use the `for name, value in ...` syntax; it'd be clearer than the `envvar[0]` syntax.

valohai_cli/commands/execution/run/dynamic_run_command.py
```diff
@@ -80,6 +80,9 @@ def __init__(self,
...
+            if envvar[0] not in self.environment_variables:
+                self.environment_variables[envvar[0]] = envvar[1].default
```
akx: Does this allow clearing default environment variables? That is, if I have something like

```yaml
environment-variables:
  foo:
    default: "hello"
```
can I clear it from being set at all with `--var foo=`?

I'd love to see a test case for that.

JuhaKiili: Was not possible today, because in our `parse_environment_variable_strings()` we had:

```
key, _, value = string.partition('=')
    if not value:
        # RAISE HELL
```

Pushed a removal of the null check and refactored the test to cover clearing the default value.

valohai_cli/commands/execution/run/dynamic_run_command.py
JuhaKiili: Fix pushed

PR #99, opened 2019-08-26
Refactor package_directory function

akx:
This adds an inner `package_files_into` function that can be used by software using `valohai-cli` as a library to further customize the contents of an ad-hoc tarball.

### DIFF
--- a/valohai_cli/adhoc.py
+++ b/valohai_cli/adhoc.py
@@ -11,7 +11,7 @@ from valohai_cli.utils.file_size_format import filesizeformat
 from valohai_cli.utils.hashing import get_fp_sha256
 
 
-def create_adhoc_commit(project, validate=True):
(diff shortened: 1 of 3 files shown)


valohai_cli/adhoc.py
```diff
@@ -51,6 +44,32 @@ def create_adhoc_commit(project, validate=True):
                 warn('Unable to remove temporary file: {}'.format(err))
 
 
+# Compatibility alias.
+# TODO: Remove in 2020.
+create_adhoc_commit = package_adhoc_commit
```
JuhaKiili: Is there some magic in Python to warn users about deprecating API?

akx: Yeah, we could wrap this with something that warns a `PendingDeprecationWarning`, but I don't think there are too many external users of the API right now.

PR #102, opened 2019-10-17
Add ignore to get_files_for_package

JuhaKiili:


valohai_cli/packager.py
```diff
@@ -106,53 +107,61 @@ def package_files_into(dest_fp, file_stats, progress=False):
...
+            for filepath
+            in files
+        ]
+
+        info('Git not available, found {n} files to package'.format(n=len(files_and_paths)))
 
     output_stats = {}
-    for file in files:
-        file_path = os.path.join(dir, file)
+    for file, file_path in files_and_paths:
```
akx: `is_valid_path()` is done in the two branches (Git and non-Git) now, but it could just be done once.

In fact, you should probably do it within this loop's body:

```python
for ...:
  if ignore and not is_valid_path(file_path, ignore):
    continue
```

so those files will get elided form the final `output_stats`.

valohai_cli/packager.py
```diff
@@ -106,53 +107,61 @@ def package_files_into(dest_fp, file_stats, progress=False):
...
 
-
-def get_files_for_package(dir, allow_git=True):
+def get_files_for_package(dir, allow_git=True, ignore=[]):
     """
     Get files to package for ad-hoc packaging from the file system.
 
     :param dir: The source directory. Probably a working copy root or similar.
     :param allow_git: Whether to allow usage of `git ls-files`, if available, for packaging.
+    :param ignore: List of ignored patterns.
```
akx: Maybe call this `ignore_patterns`.

JuhaKiili:
@akx pushed a test for the empty root folder + tested your new packager against Jupyhai and it wörks

PR #127, opened 2020-08-31
Add info of versions to development notes

magdapoppins:
As it says on the 🥫: add note about Python versions.

### DIFF
--- a/DEVELOPMENT.md
+++ b/DEVELOPMENT.md
@@ -1,5 +1,8 @@
 # Development
 
+## Setting up
+This project is written in Python 3.5 and contains tests runs in Python 3.5, 3.6 and 3.7. Create a virtualenv using Python 3.5 to get started.
+
 ```bash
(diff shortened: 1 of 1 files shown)


DEVELOPMENT.md
```diff
@@ -1,5 +1,8 @@
 # Development
 
+## Setting up
+This project is written in Python 3.5 and contains tests runs in Python 3.5, 3.6 and 3.7. Create a virtualenv using Python 3.5 to get started.
```
akx, line 4: I mean, it's smoothest to create a Py3.5 virtualenv, but the CI process will catch any non-3.5-isms anyway :)

akx (approved):

DEVELOPMENT.md
magdapoppins: It sure did catch my fstring attempt 🙃

PR #131, opened 2020-11-12
New `vh yaml` command to generate/update configs based on .py source

JuhaKiili:
Use `vh yaml step myfile.py` to generate/update the active `valohai.yaml`

### DIFF
--- a/valohai_cli/commands/yaml.py
+++ b/valohai_cli/commands/yaml.py
@@ -1,0 +1,108 @@
+import os
+from typing import Optional
+
+import click
+from valohai.internals.yaml import parse_config_from_source
+from valohai.yaml import config_to_yaml
+from valohai_yaml.objs import Config
+
+from valohai_cli.ctx import get_project
+from valohai_cli.exceptions import CLIException
+from valohai_cli.messages import info
+from valohai_cli.utils import get_project_directory
+from valohai_yaml import parse
+
+
+@click.command()
+@click.argument('filenames', nargs=-1, type=click.Path(file_okay=True, exists=True, dir_okay=False))
+def yaml(filenames):
+    """
+    Update valohai.yaml based on Python source file(s).
+
+    Example:
+
+        vh yaml hello.py
+
+    :param filenames: Path(s) of the Python source code files.
+    """
+    if not filenames:
+        raise CLIException(
+            'Missing path(s) to source Python file(s). Example: vh yaml hello.py')
+
+    project = get_project()
+    directory = (project.directory if project else get_project_directory())
+    config_path = os.path.join(directory, 'valohai.yaml')
+
+    for source_path in filenames:
+        if not os.path.isfile(config_path):
+            update_yaml_from_source(source_path, config_path)
+            info("valohai.yaml generated.")
+        elif yaml_needs_update(source_path, config_path):
+            update_yaml_from_source(source_path, config_path)
+            info("valohai.yaml updated.")
+        else:
+            info("valohai.yaml already up-to-date.")
+
+
+def get_current_config(config_path: str) -> Optional[Config]:
+    if os.path.isfile(config_path):
+        with open(config_path) as f:
+            return parse(f, validate=True)
+    return None
+
+
+def get_updated_config(source_path: str, config_path: str) -> (Config, Config):
+    """Opens the old valohai.yaml, parses source Python file and merges the resulting config to the old
+
+    Call to valohai.prepare() will contain step name, parameters and inputs.
+    We use the AST parser to parse those from the Python source code file and
+    return the merged config.
+
+    :param source_path: Path of the Python source code file
+    :param config_path: Path of the valohai.yaml config file
+
+    """
+    old_config = get_current_config(config_path)
+    new_config = parse_config_from_source(source_path, config_path)
+    if old_config:
+        return old_config.merge_with(new_config)
+    return new_config
+
+
+def update_yaml_from_source(source_path: str, config_path: str):
+    """Updates valohai.yaml by parsing the source code file for a call to valohai.prepare()
+
+    Call to valohai.prepare() will contain step name, parameters and inputs.
+    We use the AST parser to parse those from the Python source code file and
+    update (or generate) valohai.yaml accordingly.
+
+    :param source_path: Path of the Python source code file
+    :param config_path: Path of the valohai.yaml config file
+
+    """
+    old_config = get_current_config(config_path)
+    new_config = get_updated_config(source_path, config_path)
+    if old_config != new_config:
+        with open(config_path, 'w') as out_file:
+            out_file.write(config_to_yaml(new_config))
+
+
+
+def yaml_needs_update(source_path: str, config_path: str):
+    """Checks if valohai.yaml needs updating based on source Python code.
+
+    Call to valohai.prepare() will contain step name, parameters and inputs.
+    We use the AST parser to parse those from the Python source code file and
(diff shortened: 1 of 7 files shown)


valohai_cli/commands/yaml.py
```diff
@@ -0,0 +1,108 @@
...
+from valohai_cli.ctx import get_project
+from valohai_cli.exceptions import CLIException
+from valohai_cli.messages import info
+from valohai_cli.utils import get_project_directory
+from valohai_yaml import parse
+
+
+@click.command()
+@click.argument('filenames', nargs=-1, type=click.Path(file_okay=True, exists=True, dir_okay=False))
+def yaml(filenames):
```
akx, line 18: I wonder if the name here is a little loaded? Maybe make `vh yaml` a group for future expansion, and turn this into `vh yaml generate`?

valohai_cli/commands/yaml.py
```diff
@@ -0,0 +1,108 @@
...
+
+    Example:
+
+        vh yaml hello.py
+
+    :param filenames: Path(s) of the Python source code files.
+    """
+    if not filenames:
+        raise CLIException(
+            'Missing path(s) to source Python file(s). Example: vh yaml hello.py')
```
akx, line 30: You should be able to do this in the `argument` itself (`required=True` or similar)?

valohai_cli/commands/yaml.py
```diff
@@ -0,0 +1,108 @@
...
+
+    project = get_project()
+    directory = (project.directory if project else get_project_directory())
```
akx, line 33: Would just `directory = get_project_directory()` do? I didn't check the implementation so not sure.

valohai_cli/commands/yaml.py
```diff
@@ -0,0 +1,108 @@
...
+            info("valohai.yaml updated.")
+        else:
+            info("valohai.yaml already up-to-date.")
+
+
+def get_current_config(config_path: str) -> Optional[Config]:
+    if os.path.isfile(config_path):
+        with open(config_path) as f:
+            return parse(f, validate=True)
+    return None
```
akx, line 51: I would imagine this utility exists in valohai-cli already...

valohai_cli/commands/yaml.py
```diff
@@ -0,0 +1,108 @@
...
+
+    for source_path in filenames:
+        if not os.path.isfile(config_path):
+            update_yaml_from_source(source_path, config_path)
+            info("valohai.yaml generated.")
+        elif yaml_needs_update(source_path, config_path):
+            update_yaml_from_source(source_path, config_path)
+            info("valohai.yaml updated.")
+        else:
+            info("valohai.yaml already up-to-date.")
```
akx, line 44: Won't this just happily overwrite `config_path` for each `source_path`?

akx (changes requested):
Sorry for the super late review 😞

valohai_cli/commands/yaml.py
JuhaKiili: For example `vh lint` uses the same pattern that I'm doing here. https://github.com/valohai/valohai-cli/blob/988c7a2b415ad244379ce8af4e4ca0ff74ea5b36/valohai_cli/commands/lint.py#L47

`get_project_directory()` has simpler logic than `get_project().directory`. To be honest, I don't quite understand why there are these two ways, but I took the `vh lint` as the ground truth as it is trying to find the correct `valohai.yaml` just like my command.

JuhaKiili: Edit: I refactored the code so that I'm using the Project object and don't need this code anymore

valohai_cli/commands/yaml.py
JuhaKiili: No. For each `source_path` the same `valohai.yaml` is updated, **not** overwritten.

If you feed this command multiple `source_path`, each will contain a different Valohai step, and updates a different part of the config file.

valohai_cli/commands/yaml.py
JuhaKiili: Pushed fix

valohai_cli/commands/yaml.py
JuhaKiili: Changed to `vh yaml step foobar.py`

The reasoning is that we only generate/update single step here. We might have things like `vh yaml endpoint` or `vh yaml pipeline`.

valohai_cli/commands/yaml.py
JuhaKiili: Pushed fix

valohai_cli/commands/yaml/step.py
```diff
@@ -0,0 +1,97 @@
...
+            info("valohai.yaml updated.")
+        else:
+            info("valohai.yaml already up-to-date.")
+
+
+def get_current_config(project: Project) -> Optional[Config]:
+    return project.get_config() if os.path.isfile(project.get_config_filename()) else None
+
+
+def get_updated_config(source_path: str, project: Project) -> (Config, Config):
```
akx, line 44: Shouldn't the rtype be

```suggestion
def get_updated_config(source_path: str, project: Project) -> Config:
```

valohai_cli/commands/yaml/step.py
```diff
@@ -0,0 +1,97 @@
...
+
+    """
+    old_config = get_current_config(project)
+    new_config = parse_config_from_source(source_path, project.get_config_filename())
+    if old_config:
+        return old_config.merge_with(new_config)
+    return new_config
+
+
+def update_yaml_from_source(source_path: str, project: Project):
```
akx, line 62: Wouldn't hurt to return a "did something" flag from here :)
```suggestion
def update_yaml_from_source(source_path: str, project: Project) -> bool:
```

valohai_cli/commands/yaml/step.py
```diff
@@ -0,0 +1,97 @@
...
+    :param source_path: Path of the Python source code file
+    :param project: Currently linked Valohai project
+
+    """
+    old_config = get_current_config(project)
+    new_config = get_updated_config(source_path, project)
+    if old_config != new_config:
+        with open(project.get_config_filename(), 'w') as out_file:
+            out_file.write(config_to_yaml(new_config))
+
```
akx, line 78:
```suggestion
            out_file.write(config_to_yaml(new_config))
        return True
    return False
```

valohai_cli/commands/yaml/step.py
```diff
@@ -0,0 +1,97 @@
...
+
+def yaml_needs_update(source_path: str, project: Project):
```
akx, line 80:
```suggestion
def yaml_needs_update(source_path: str, project: Project) -> bool:
```

valohai_cli/commands/yaml/step.py
```diff
@@ -0,0 +1,97 @@
...
+            info("valohai.yaml generated.")
+        elif yaml_needs_update(source_path, project):
+            update_yaml_from_source(source_path, project)
+            info("valohai.yaml updated.")
+        else:
+            info("valohai.yaml already up-to-date.")
+
+
+def get_current_config(project: Project) -> Optional[Config]:
+    return project.get_config() if os.path.isfile(project.get_config_filename()) else None
```
akx, line 41: Wonder if this'd be better phrased as

```suggestion
def get_current_config(project: Project) -> Optional[Config]:
    try:
        return project.get_config()
    except FileNotFoundError:
        return None
```

akx (approved):
🚢

Issue #64, opened 2019-03-29
Stop supporting Python 2


akx:
There's no good reason to support Python 2, as [support for it will end in 9 months at the time of writing](https://pythonclock.org/).

In addition, Python 3 versions older than 3.5 are also end-of-life, so 3.5+ should be our target.

freducom:
When discontinuing py2 support make sure there is some clear error for people that still try to get it to work with py2.

PR #134, opened 2021-02-03
Run Pyupgrade, drop six dependency

akx:
Some of these changes could have been done around #64 / #106, but better late than never, I guess...


magdapoppins (approved):
🏁

PR #138, opened 2021-03-02
Run flynt to f-stringify things

akx:
Now that we're no longer required to support Python 3.5, we can go all in with f-strings.


magdapoppins (approved):
Seems thoroughly f-stringified. 🏁

PR #139, opened 2021-03-02
Type annotations

akx:
Provided by Monkeytype and Mypy, lovingly massaged by hand...


magdapoppins (approved):
Skimming through: looks good 💎

PR #141, opened 2021-04-08
Add requirements.txt when creating a new yaml step

DrazenDodik:
Add ``valohai-utils`` to an existing requirements.txt or create a new requirements.txt with ``valohai-utils``


akx:
Fixed up in 666427bad51c2809c00b5efee6dad42169e6264c (also reverted the earlier merge of #140).

PR #142, opened 2021-04-14
Add vh yaml pipeline command

JuhaKiili:
Allow users to create/update a pipeline in their `valohai.yaml`, using the following command:

`vh yaml pipeline mypipeline.py`

The actual logic for getting the Pipeline config is not here, but in `valohai.internals.pipeline.get_pipeline_from_source()`.


setup.py
```diff
@@ -12,7 +12,7 @@
     install_requires=[
         'click>=7.0',
         'valohai-yaml>=0.9',
-        'valohai-utils>=0.1.2',
+        'valohai-utils>=0.1.7',
```
akx, line 15: The highest version of `valohai-utils` on https://github.com/valohai/valohai-utils/tree/699d352a39251ddc4b195e039ae807a1ccdba740 is `0.1.6`. Did you forget to push 0.1.7 to git, or are there other shenanigans going on here?

akx (changes requested):

setup.py
JuhaKiili: My bad. Forgot to push indeed.

akx (approved):

PR #147, opened 2021-06-01
Replace grave accents with quotes

orasimus:
Replace grave accents with quotes in printed help, as it causes issues with reStructuredText used to generate Valohai's official documentation.


akx (approved):
A grave mistake heroically mended.

PR #152, opened 2021-06-09
Embetter login host prompt flow

### LINKED ISSUE #151: Ask for Valohai host (with the default being the default) when logging in interactively

akx:
With this PR:

* login will ask to verify the login host; one can just hit ENTER to accept the default (app.valohai.com, or whatever had been overridden using the `VALOHAI_HOST` envvar)
* login will not fail if the host (either via arg or interactive prompt) is malformed (not a real base URL)

Fixes #151

# What does it look like?

## default interactive flow + validation

Screenshot 2021-06-09 at 16 55 01

## host defined on command line

Screenshot 2021-06-09 at 16 55 29

### DIFF
--- a/tests/commands/test_auth.py
+++ b/tests/commands/test_auth.py
@@ -38,7 +38,11 @@ def test_auth(runner, mode):
                 '-p', m.password,
             ], catch_exceptions=False)
         elif mode == 'interactive':
-            result = runner.invoke(login, input=f'{m.username}\n{m.password}')
+            result = runner.invoke(login, input='\n'.join([
+                '',  # accept default host
+                m.username,  # enter username
+                m.password,  # enter password
+            ]))
         assert 'Logged in' in result.output
         assert settings.token == m.token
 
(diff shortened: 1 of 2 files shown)


ruksi (changes requested):
improving the login flow to be more intuitive

lgtm

`mypy` says nono on some type hints though

tokkoro:
Does this login me into myhai.com?
`vh --valohai-host https://myhai.com/ login`

akx:
> Does this login me into myhai.com?
> `vh --valohai-host https://myhai.com/ login`

Yes, it asks you whether to log in on https://myhai.com/ or whether to override it.

tokkoro:
:+1: super

tokkoro (approved):
I'm not totally sure why an user would need to type https:// in front of host name, it could be a default and can be overwritten with http:// if explicitly typed. That would be more :potato:proof.

akx:
> `mypy` says nono on some type hints though

Mypy is calm now.

ruksi (approved):
lgtm

> I'm not totally sure why an user would need to type https:// in front of host name, it could be a default and can be overwritten with http:// if explicitly typed. That would be more proof.

yeah, could be a nice addition, I have nothing against that

I think this is a good first version as is and that could be added later?

PR #155, opened 2021-07-05
Package files from submodules too

akx:
Cc @johncharrington, who's using a variation of this code right now :)

### DIFF
--- a/valohai_cli/packager.py
+++ b/valohai_cli/packager.py
@@ -113,11 +113,27 @@ def package_files_into(
 
 
 def _get_files_with_git(dir: str) -> Iterable[Tuple[str, str]]:
-    for line in check_output('git ls-files --exclude-standard -ocz', cwd=dir, shell=True).split(b'\0'):
-        if line.startswith(b'.'):
-            continue
-        file = line.decode('utf-8')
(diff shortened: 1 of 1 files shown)


valohai_cli/packager.py
```diff
@@ -113,11 +113,27 @@ def package_files_into(
...
+    commands = [
+        'git ls-files --exclude-standard -ocz',
+    ]
+
+    # Check whether a gitmodules file exists; if so, also do a `--recurse-submodules` pass
+    # to gather files in submodules.  Untracked files in submodules will not be considered,
+    # since `--recurse-submodules` does not support `-o`.
+    gitmodules_file = os.path.join(dir, '.gitmodules')
+    if os.path.isfile(gitmodules_file):
+        commands.append('git ls-files --exclude-standard --recurse-submodules -cz')
```
ruksi, line 126:
```
--recurse-submodules:
	Recursively calls ls-files on each submodule in the repository.
	Currently there is only support for the --cached mode.
```

sounds that it does what it says on the tin, and also the `-o` redacted as it only supports cached files (tracked)

ruksi (approved):
looks legit

Issue #156, opened 2021-07-09
Update error message for creating pipelines with utils


noorula:
I was trying to create a pipeline in the CLI with valohai-utils and kept receiving an error message saying that a step with the specified name doesn't exist in my yaml. After banging my head on the wall for some time I realized to check which project is connected to that folder and apparently it was not the one I thought it was. 

To prevent further violent wall+head meetings, could be update the error message to contain something like:

"Run vh project status to see which project is your current directory linked to" ?

akx:
The error message could be updated to show the filename of the YAML it's inspecting.

PR #158, opened 2021-07-26
Pipeline fixes

akx:
This PR:

* fixes some miscellaneous error messages in the pipeline... pipeline
* makes it so that the pipeline-generating code is only executed once per file (my previous pull request review hadn't taken that into consideration)
* catches exceptions from the pipeline-generating code to prefix them with some context (pending valohai/papi#9 to make the errors even better), see below. (#156)

## example contextful output

```
$ vh yaml pipe xx.py
😢  ERROR: Retrieving a new pipeline definition for project asdf-pipeline-demo for xx.py failed.
The configuration file in use is /Users/akx/build/tensorflow-example/valohai.yaml. See the full traceback below.
Traceback (most recent call last):
  [...]
  File "xx.py", line 8, in main
    extract = pipe.execution("Batch feature extraction")
  File "papi/__init__.py", line 40, in execution
    step_object = self._get_step(step)
  File "papi/__init__.py", line 50, in _get_step
    raise ValueError(f"no such step {name} in config")
ValueError: no such step Batch feature extraction in config
```

### DIFF
--- a/valohai_cli/commands/yaml/pipeline.py
+++ b/valohai_cli/commands/yaml/pipeline.py
@@ -6,6 +6,7 @@ from valohai.yaml import config_to_yaml
 from valohai_yaml.objs.config import Config
 
 from valohai_cli.ctx import get_project
+from valohai_cli.exceptions import ConfigurationError
 from valohai_cli.messages import info
 from valohai_cli.models.project import Project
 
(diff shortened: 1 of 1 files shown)


magdapoppins (approved):
Very nice improvements! 💎

PR #163, opened 2021-09-13
Add limited support for .gitignore even when git is not installed

### LINKED ISSUE #162: Adhoc executions can
Adhoc executions don't ignore files and folders from a .gitignore file

JuhaKiili:
When new users are onboarded or new projects are created, installing git and initializing repository, etc. are just more things to remember. We want to give the users a frictionless path to their first executions. 

That is why we are adding our own fallback support for `.gitignore`. It offers a step in the right direction (using the `.gitignore` file), without the extra steps of installing git or running `git init`.

I'll admit that using `.gitignore` without `git` is not very elegant, but it makes sense here.

Fixes #162


valohai_cli/packager.py
```diff
@@ -148,14 +155,24 @@ def _get_files_walk(dir: str) -> Iterable[Tuple[str, str]]:
...
+            try:
+                return (GitUsage.GIT_LS_FILES, _get_files_with_git(dir))  # return the generator
+            except subprocess.CalledProcessError as cpe:
+                warn(f'.git exists, but we could not use git ls-files (error {cpe.returncode}), falling back to non-git')
+
+        # Limited support of .gitignore even without git
+        elif os.path.exists(gitignore_path):
+            with open(gitignore_path, "r") as git_ignore_file:
+                ignore_patterns = [pattern.strip() for pattern in git_ignore_file.readlines()]
+                return (GitUsage.GITIGNORE_WITHOUT_GIT, [p for p in _get_files_walk(dir) if is_valid_path(p[0], ignore_patterns)])
```
akx, line 173: This should return a genexpr, not a list:

```suggestion
                return (GitUsage.GITIGNORE_WITHOUT_GIT, (p for p in _get_files_walk(dir) if is_valid_path(p[0], ignore_patterns)))
```

valohai_cli/packager.py
```diff
@@ -191,10 +208,12 @@ def get_files_for_package(
                     threshold=FILE_COUNT_HARD_THRESHOLD,
                 ))
 
-    if using_git:
+    if git_usage == GitUsage.GIT_LS_FILES:
         info(f'Used git to find {len(files_and_paths)} files to package')
+    elif git_usage == GitUsage.GITIGNORE_WITHOUT_GIT:
+        info(f'Used .gitignore to find {len(files_and_paths)} files to package')
```
akx, line 214: Should this note that we're limited to a, well, limited form of `.gitignore` parsing? 

I mean, the format isn't exactly described in one line only, and I'm almost sure we don't support all of the features... https://git-scm.com/docs/gitignore

akx (changes requested):

valohai_cli/packager.py
```diff
@@ -40,6 +41,12 @@
 PackageFileInfo = namedtuple('PackageFileInfo', ('source_path', 'stat'))
 
 
+class GitUsage(Enum):
+    NONE = 0,
+    GITIGNORE_WITHOUT_GIT = 1,
+    GIT_LS_FILES = 2,
```
akx, line 47: I doubt you wanted to make these 1-tuples? 😄 
```suggestion
    NONE = 0
    GITIGNORE_WITHOUT_GIT = 1
    GIT_LS_FILES = 2
```

valohai_cli/packager.py
```diff
@@ -148,14 +155,24 @@ def _get_files_walk(dir: str) -> Iterable[Tuple[str, str]]:
...
+    if allow_git:
+        if os.path.exists(os.path.join(dir, '.git')):
+            # We have .git, so we can try to use Git to figure out a file list of nonignored files
+            try:
+                return (GitUsage.GIT_LS_FILES, _get_files_with_git(dir))  # return the generator
+            except subprocess.CalledProcessError as cpe:
+                warn(f'.git exists, but we could not use git ls-files (error {cpe.returncode}), falling back to non-git')
+
+        # Limited support of .gitignore even without git
+        elif os.path.exists(gitignore_path):
```
akx, line 170: This should probably be `if`? That way the `warn()` from above falls back to our gitignore emulation when possible?
```suggestion
        if os.path.exists(gitignore_path):
```

valohai_cli/packager.py
```diff
@@ -148,14 +155,24 @@ def _get_files_walk(dir: str) -> Iterable[Tuple[str, str]]:
...
+            with open(gitignore_path, "r") as git_ignore_file:
+                ignore_patterns = [pattern.strip() for pattern in git_ignore_file.readlines()]
+                return (GitUsage.GITIGNORE_WITHOUT_GIT, (p for p in _get_files_walk(dir) if is_valid_path(p[0], ignore_patterns)))
```
akx, line 173: No need to keep the file open during this time.
```suggestion
            return (GitUsage.GITIGNORE_WITHOUT_GIT, (p for p in _get_files_walk(dir) if is_valid_path(p[0], ignore_patterns)))
```

akx (changes requested):
Sorry, a couple more things I'd missed the first time around...

akx (approved):

PR #165, opened 2021-09-16
Notify the user if a project is missing in vh yaml

magdapoppins:
Fixes valohai/valohai-utils#64

### DIFF
--- a/valohai_cli/commands/yaml/step.py
+++ b/valohai_cli/commands/yaml/step.py
@@ -25,7 +25,11 @@ def step(filenames: List[str]) -> None:
     :param filenames: Path(s) of the Python source code files.
     """
     project = get_project()
-    assert project
+    if project is None:
+        info("no project linked - assuming files are in current directory.")
+        project = Project(data={}, directory=os.getcwd())
+        project.name = "YAML command simulated project"
+
     config_path = project.get_config_filename()
 
     for source_path in filenames:


akx (commented):
What is the assertion error thrown by valohai_utils? Could we fix the command to work without a project? I see no reason why YAML generation would require a linked project.

magdapoppins:
> What is the assertion error thrown by valohai_utils? Could we fix the command to work without a project? I see no reason why YAML generation would require a linked project.

Requires some decoupling, but should be doable!

valohai_cli/commands/yaml/step.py
```diff
@@ -25,8 +26,11 @@ def step(filenames: List[str]) -> None:
     :param filenames: Path(s) of the Python source code files.
     """
     project = get_project()
-    assert project
-    config_path = project.get_config_filename()
+
+    if project is None:
+        info("no project linked - assuming files are in current directory.")
```
akx, line 31: Can you try if this works with just

```suggestion
    if project is None:
        info("no project linked - assuming files are in current directory.")
        project = Project(data={}, directory=os.getcwd())
```

i.e. you simulate a project that would live in this directory?

That way you don't need to change everything to accept no Project.

akx (changes requested):

akx (approved):
Nice and clean 😁

PR #167, opened 2021-09-28
Allow disabling SSL verification when logging in

### LINKED ISSUE #161: Enable logging to "insecure" Valohai instances
Currently the `-h/--host` parameter doesn't work if the Valohai host instance doesn't operate under a valid TLS certificate address.

Depending on the host installation, sometimes they might not have a certificate or they might have for example a self-signed certificate.

akx:
Fixes #161


ruksi (approved):
less security if requested

lgtm

PR #168, opened 2021-09-28
Version bumps & type fixes

akx:
This PR:

* bumps the required minimum versions of the valohai-* packages to somewhat newer versions
* increases test compatibility with newer versions of Click that have changed some spellings
* fixes some type errors found by `mypy --strict`


ruksi (approved):
dependency bumps and small refactorings

lgtm

PR #180, opened 2021-11-09
Add transient setting `api_user_agent_prefix`

akx:
This allows downstream library users to modify the user-agent reported by valohai-cli without having to override, like, everything.


ruksi (approved):
allow modifying the user agent that `valohai-cli` reports forward

lgtm!

PR #181, opened 2021-12-21
Use valohai-yaml for pipeline payload generation

akx:
Following https://github.com/valohai/valohai-yaml/pull/61, valohai-yaml has the canonical code to convert a pipeline definition to a payload.


tokkoro (approved):
It's always nice to see a ratio of 9 removed lines per 1 new addition :)

PR #182, opened 2021-12-21
Declarative setup + tool upgrades

akx:
á la https://github.com/valohai/hai/pull/20 & https://github.com/valohai/valohai-yaml/pull/62 ...


ruksi (approved):
package definitions from `cfg` instead from `py` and upgrades

lgtm!

PR #190, opened 2022-04-14
Use images.v2.yaml instead of legacy images.yaml

akx:
Refs valohai/images#4


ruksi (changes requested):
`mypy` not happy

ruksi (approved):
lgtm

PR #192, opened 2022-05-03
Allow creating adhoc commits with a nonstandard yaml path

magdapoppins:
- When creating an adhoc commit, ask roi what the yaml path of that project is and send that path over to `import_package` in roi

Fixes valohai/roi#3953

### DIFF
--- a/valohai_cli/commands/execution/run/frontend_command.py
+++ b/valohai_cli/commands/execution/run/frontend_command.py
@@ -66,6 +66,7 @@ def run(
         ctx.exit()
         return
     project = get_project(require=True)
+    project.refresh_details()
 
     if download_directory and watch:
         raise click.UsageError('Combining --sync and --watch not supported yet.')
--- a/valohai_cli/commands/lint.py
+++ b/valohai_cli/commands/lint.py
@@ -43,8 +43,13 @@ def lint(filenames: List[str]) -> None:
     """
     if not filenames:
         project = get_project()
+        if project:
+            project.refresh_details()
+            yaml_path = project.get_yaml_path()
+        else:
+            yaml_path = 'valohai.yaml'
         directory = (project.directory if project else get_project_directory())
-        config_file = os.path.join(directory, 'valohai.yaml')
+        config_file = os.path.join(directory, yaml_path)
         if not os.path.exists(config_file):
             raise CLIException(f'There is no {config_file} file. Pass in the names of configuration files to lint?')
         filenames = [config_file]
--- a/valohai_cli/models/project.py
+++ b/valohai_cli/models/project.py
@@ -60,7 +60,11 @@ class Project:
             ))
 
     def get_config_filename(self) -> str:
-        return os.path.join(self.directory, 'valohai.yaml')
+        return os.path.join(self.directory, self.get_yaml_path())
+
+    def get_yaml_path(self) -> str:
+        # Make sure the older API versions that don't expose yaml_path don't completely break
+        return self.data.get('yaml_path', 'valohai.yaml')
 
     def get_execution_from_counter(
         self,
@@ -138,6 +142,16 @@ class Project:
                 return data
 
         raise ValueError(f'No commit found for commit {identifier}')
+
+    def refresh_details(self) -> None:
+        """
+        Refresh the project details from the API.
+        """
+        data = request(
+            'get',
+            f'/api/v0/projects/{self.id}/',
+        ).json()
+        self.data.update(data)
 
     def __str__(self) -> str:
         return self.name
--- a/valohai_cli/packager.py
+++ b/valohai_cli/packager.py
@@ -53,11 +53,11 @@ class VhIgnoreUsage(Enum):
     VHIGNORE = 1
 
 
-def package_directory(dir: str, progress: bool = False, validate: bool = True) -> str:
-    file_stats = get_files_for_package(dir)
+def package_directory(*, directory: str, yaml_path: str, progress: bool = False, validate: bool = True) -> str:
+    file_stats = get_files_for_package(directory)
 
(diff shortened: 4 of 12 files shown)


valohai_cli/commands/project/link.py
```diff
@@ -33,7 +33,7 @@ def choose_project(dir: str, spec: Optional[str] = None) -> Optional[dict]:
     :param spec: An optional search string
     :return: project object or None
     """
-    projects: List[dict] = request('get', '/api/v0/projects/', params={'limit': '1000'}).json()['results']
+    projects: List[dict] = request('get', '/api/v0/projects/', params={'limit': '1000', 'include': 'yaml_path'}).json()['results']
```
akx, line 36: Hmm... this will not pick up changes to the yaml path that are done after linking the project. :(

May need to discuss how to best do this in meatspace :)

valohai_cli/commands/lint.py
```diff
@@ -44,7 +44,7 @@ def lint(filenames: List[str]) -> None:
     if not filenames:
         project = get_project()
         directory = (project.directory if project else get_project_directory())
-        config_file = os.path.join(directory, 'valohai.yaml')
+        config_file = os.path.join(directory, project.data.get('yaml_path', 'valohai.yaml') if project else 'valohai.yaml')
```
akx, line 47: Maybe the `project.data.get('yaml_path', 'valohai.yaml') if project else 'valohai.yaml'` expression should be broken out to a separate line for clarity?

Also, I suppose it should use `project.get_config_filename()`?

valohai_cli/models/project.py
```diff
@@ -60,7 +60,8 @@ def _parse_config(self, config_fp: TextIO, filename: str = '<config file>') -> C
             ))
 
     def get_config_filename(self) -> str:
-        return os.path.join(self.directory, 'valohai.yaml')
+        filename = self.data['yaml_path'] if self.data.get('yaml_path') else 'valohai.yaml'
```
akx, line 63:
```suggestion
        filename = self.data.get('yaml_path') or 'valohai.yaml'
```

akx (changes requested):
Beside the inline comments... I think we'd need to post the YAML path with the adhoc upload endpoint?

valohai_cli/models/project.py
```diff
@@ -60,7 +60,12 @@ def _parse_config(self, config_fp: TextIO, filename: str = '<config file>') -> C
             ))
 
     def get_config_filename(self) -> str:
-        return os.path.join(self.directory, 'valohai.yaml')
+        project_details: dict = request(
+            'get',
+            f'/api/v0/projects/{self.id}/',
+        ).json()
+        yaml_path = project_details['yaml_path']
+        return os.path.join(self.directory, yaml_path)
```
akx, line 68: Hmm... I'm not sure it's a good idea that we implicitly do a request every time `get_config_filename` is called.

Since we do cache some project info locally anyway, what if...

* add a `refresh_details()` function that we call at the start of `run` or `lint`
* it saves the project detail information in the local persistence
* a `Project.get_yaml_path()` function gets the yaml path from the local persistence
* this would use that (unless the YAML path is explicitly overridden for e.g. adhoc runs)

?

akx (changes requested):

valohai_cli/packager.py
```diff
@@ -53,11 +53,11 @@ class VhIgnoreUsage(Enum):
...
 
-def package_directory(dir: str, progress: bool = False, validate: bool = True) -> str:
-    file_stats = get_files_for_package(dir)
+def package_directory(*, directory: str, yaml_path: str, progress: bool = False, validate: bool = True) -> str:
+    file_stats = get_files_for_package(directory)
 
-    if validate and 'valohai.yaml' not in file_stats:
-        raise ConfigurationError(f'valohai.yaml missing from {dir}')
+    if validate and yaml_path not in file_stats:
+        raise ConfigurationError(f'valohai.yaml missing from {directory}')
```
akx, line 60: Should this message also refer to the YAML path specified?

valohai_cli/models/project.py
```diff
@@ -60,7 +60,21 @@ def _parse_config(self, config_fp: TextIO, filename: str = '<config file>') -> C
...
+
+    def refresh_details(self) -> None:
+        """
+        Refresh the project details from the API.
+        """
+        data = request(
+            'get',
+            f'/api/v0/projects/{self.id}/',
+        ).json()
+        self.data.update(data)
```
akx, line 73: For prettiness's sake, maybe this function shouldn't be sandwiched between the two related `get_*` functions, but somewhere else in the class?

valohai_cli/commands/lint.py
```diff
@@ -43,8 +43,11 @@ def lint(filenames: List[str]) -> None:
     """
     if not filenames:
         project = get_project()
+        if project:
+            project.refresh_details()
+        yaml_path = project.get_yaml_path() if project else 'valohai.yaml'
```
akx, line 48:
```suggestion
        if project:
            project.refresh_details()
            yaml_path = project.get_yaml_path()
        else:
            yaml_path = 'valohai.yaml'
```

valohai_cli/commands/execution/run/frontend_command.py
```diff
@@ -66,6 +66,7 @@ def run(
         ctx.exit()
         return
     project = get_project(require=True)
+    project.refresh_details()
```
akx, line 69: Somewhat unrelated to this line, but didn't we talk about the run command (especially in adhoc mode) to be able to accept a YAML path argument? Should that be in this PR, or would it be in a forthcoming one?

magdapoppins: Broke this into #193 ✅

PR #200, opened 2022-06-20
Add data and alias command to fetch Datums and Datum Aliases

tokkoro:
fixes [#196](https://github.com/valohai/valohai-cli/issues/196)

# Valohai-cli Data

## Data
In a linked project directory, you can read datums using `vh data list` or from specific project by `vh --project <project-id> data list`

**Columns**
* Name: The name of a datum 
* Size: The file size of a datum
* Output of Exec: The execution that a datum came from
* Created At: The time when a datum was created
* URL: datum:// url to use as an input for execution for example
* URI: Storage uri for the file that a datum is pointing.

### Example output
```text
Name                        Size     Output of Exec  Created At                   URL                                           URI                                                                                    
--------------------------  -------  --------------  ---------------------------  --------------------------------------------  ----------------------------------------------------------
sweed_music_of_birds.wav     9.0 MB  Not from exec   2022-05-19T07:46:31.696250Z  datum://0180db51-7a50-679f-d635-cafb925c2122  s3://bucket/data/PATH/upload/sweed_music_of_birds.wav
small_data_file_ouput.txt   18.0  B  #756            2022-05-15T11:12:35.408585Z  datum://0180c735-8630-9812-b08e-cf1f2556d10c  s3://bucket/data/PATH/output-756/small_data_file_ouput.txt        
output_file.txt             18.0  B  #752            2022-05-14T10:13:26.512045Z  datum://0180c735-636f-a104-33b4-1972a83a492b  s3://bucket/data/PATH/output-752/output_file.txt
```

## Alias
In a linked project directory, you can read datum aliases using `vh alias list` or from specific project by `vh --project <project-id> alias list`

**Columns**
* Name: The name of an alias 
* Datum: The name of the datum that an alias is pointing to.
* Last Modified: The time an alias was previously updated
* URL: datum:// url to use as an input for execution for example
### Example output
```text
Name     Datum                Last Modified                URL            
-------  -------------------  ---------------------------  ---------------
most-ac  layer-0-weights.csv  2022-06-17T10:22:40.679929Z  datum://most-ac
alias-1  layer-1-biases.csv   2022-06-17T10:22:31.754594Z  datum://alias-1
the-eht  layer-1-biases.csv   2022-06-17T10:22:24.286183Z  datum://the-eht
alias-0  layer-0-weights.csv  2022-06-17T10:21:09.323021Z  datum://alias-0

```

### DIFF
--- a/valohai_cli/commands/alias/list.py
+++ b/valohai_cli/commands/alias/list.py
@@ -1,0 +1,48 @@
+import click
+
+from valohai_cli.api import request
+from valohai_cli.ctx import get_project
+from valohai_cli.messages import info
+from valohai_cli.settings import settings
+from valohai_cli.table import print_json, print_table
+
+
+@click.command()
+@click.option(
+    '--count',
+    '--limit',
+    '-n',
+    type=int,
+    default=100,
+    help='How many aliases to show',
+)
+def list(count: int) -> None:
+    """
+    Show a list of data aliases in the project.
+    """
+    params = {
+        'limit': count,
+        'ordering': '-ctime',
+        'deleted': 'false',
+        'no_count': 'true'
+    }
+    project = get_project(require=True)
+    assert project
+    if project:
+        params['project'] = project.id
+
+    aliases = request('get', '/api/v0/datum-aliases/', params=params).json()['results']
+    if settings.output_format == 'json':
+        return print_json(aliases)
+    if not aliases:
+        info(f'{project}: No data aliases.')
+        return
+    for alias in aliases:
+        alias['url'] = f'datum://{alias["name"]}'
+        alias['datum'] = 'No target' if not alias['datum'] else alias['datum']['name']
+
+    print_table(
+        aliases,
+        columns=['name', 'datum', 'mtime', 'url'],
+        headers=['Name', 'Data', 'Last Modified', 'URL'],
+    )
--- a/valohai_cli/commands/data/__init__.py
+++ b/valohai_cli/commands/data/__init__.py
@@ -1,0 +1,10 @@
+import click
+
+from valohai_cli.plugin_cli import PluginCLI
+
+
+@click.command(cls=PluginCLI, commands_module='valohai_cli.commands.data')
+def data() -> None:
+    """
+    Data-related commands.
+    """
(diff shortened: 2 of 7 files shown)


valohai_cli/commands/alias/list.py
```diff
@@ -0,0 +1,48 @@
...
+    '--limit',
+    '-n',
+    type=int,
+    default=100,
+    help='How many datums to show',
+)
+def list(count: int) -> None:
+    """
+    Show a list of datums for the project.
+    """
```
ruksi, line 22: a list of datum aliases? probably a copy-paste error

valohai_cli/commands/data/__init__.py
```diff
@@ -0,0 +1,10 @@
+import click
+
+from valohai_cli.plugin_cli import PluginCLI
+
+
+@click.command(cls=PluginCLI, commands_module='valohai_cli.commands.data')
+def data() -> None:
+    """
+    Datum-related commands.
+    """
```
ruksi, line 10: slightly off-put but mixing three terms: `data`, `datums` and `datum`

there should just be two terms; singular and plural
* `datum` & `datums`
* `datum` & `data`

so the `data` is only the customer facing term here? should this be reflected e.g. in the documentation and the web app guide texts?

not a major issue but it makes even the reviewing slightly more troublesome, I can imagine it can make the usage without code completion even more annoying

maybe we should give this a minor discussion during next daily, but not a blocking issue if the change is otherwise deemed essential

ruksi (changes requested):
let's huddle on this this week :+1:

valohai_cli/commands/data/__init__.py
tokkoro: I fixed this so that the user interface is always talking about **data**. Most of the code is also, but for example test data is named as custom in `fixture_data.py`.

ruksi (approved):
lgtm!

`convert_size` could be moved into some utility directory for reuse, but this can be done later if the size conversions are needed later (moving it now could improve discoverability)

there could be more options for the `list` commands like customizing sort etc. but this is an excellent first version

PR #201, opened 2022-06-20
Default to valohai.yaml properly when path in project data is None

magdapoppins:
When creating a new project via `vh init` the project doesn't have a connected repository in valohai, so the yaml path in the project data is also set as `None`. Let's always make it default to `valohai.yaml`.


orasimus (approved):
lgtm

PR #202, opened 2022-06-22
Keep create_adhoc_commit_from_tarball in tact

### LINKED ISSUE #194: Ensure custom YAML paths don't interfere with Jupyhai's usage of valohai-cli
Jupyhai uses valohai-cli as an API driver, so we'll need to make sure it doesn't break (e.g. inadvertently always use a wrong valohai.yaml?) when #192 is released.

magdapoppins:
Since `create_adhoc_commit_from_tarball` is being used in jupyhai which isn't aware of custom yaml paths per command, only make yaml_path mandatory in the following private `_upload_commit_code` function and use the project settings yaml as default.

(Also came across a comment to delete a compatibility alias in 2020 so I did.)

Fixes #194

### DIFF
--- a/valohai_cli/adhoc.py
+++ b/valohai_cli/adhoc.py
@@ -48,9 +48,6 @@ def package_adhoc_commit(project: Project, validate: bool = True, yaml_path: Opt
                 warn(f'Unable to remove temporary file: {err}')
 
 
-# Compatibility alias.
-# TODO: Remove in 2020.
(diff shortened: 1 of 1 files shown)


ruksi (approved):
autoresolve `yaml_path` in `create_adhoc_commit_from_tarball ` if not defined so API driver usage works properly e.g. for Jupyhai

lgtm!

PR #203, opened 2022-06-23
Add description values to package metadata

ruksi:
These changes will make the PyPI page more helpful after the next release.

The current information is pretty slim:

Screenshot from 2022-06-23 11-35-06


magdapoppins (approved):
Nice!

PR #213, opened 2022-09-19
Add --open-browser option to `exec run`

akx:
It's like `--watch`, but with more WWW!


valohai_cli/commands/execution/run/dynamic_run_command.py
```diff
@@ -81,6 +82,7 @@ def __init__(
         self.environment = environment
         self.image = image
         self.watch = bool(watch)
+        self.open_browser = bool(open_browser)
```
ruksi, line 85: not updated to the `RunCommand.__init__` docstring but it seems to be a bit of a mess anyway

might require a cleanup or removal, possibly in a separate PR

ruksi (approved):
lgtm!

PR #214, opened 2022-09-29
Allow ad-hoc deployment version creation

magdapoppins:
If the `--adhoc` flag is passed, similarly to the run execution command, create a new adhoc commit of the current working directory and use it to create a new deployment version.


akx (approved):
LGTM (as soon as the Roi stuff for this is live).

Issue #216, opened 2022-10-14
Upgrading valohai-cli does not update valohai-yaml


noorula:
I upgraded `valohai-cli` but kept getting error for pipeline parameters. Manually updating `valohai-yaml` solved this but from user perspective it would be better if there was only one component to update.

akx:
You can run `pip install -U --upgrade-strategy=eager valohai-cli` to have pip upgrade `valohai-cli` and all dependencies too. But sure, no harm in bumping the minimum required `valohai-yaml` version in `valohai-cli` too :)

teroyks:
@noorula This issue can probably be closed?

noorula:
Sure thing! @teroyks

PR #217, opened 2022-10-14
Bump valohai* deps to their current freshest versions

akx:
Refs #216


magdapoppins (approved):
Might we want to add upgrades within the `.in` constraints as an automatic step of releasing a new version?

akx:
> Might we want to add upgrades within the `.in` constraints as an automatic step of releasing a new version?

https://github.com/valohai/valohai-cli/pull/220 should do that.

ruksi (approved):
looking fresh :green_salad:

PR #218, opened 2022-10-14
Validate commit in respect to the active project

### LINKED ISSUE #104: CLI gives API error on running cached ad-hoc commit from a different project
Screenshot from 2019-10-29 13-45-05

Repro steps:
1. Run ad-hoc execution
2. Run it again (works)
3. Create & link new project `vh project create`
4. Run the execution again
5. API error occurs

magdapoppins:
Fixes #104

### DIFF
--- a/valohai_cli/adhoc.py
+++ b/valohai_cli/adhoc.py
@@ -59,7 +59,7 @@ def create_adhoc_commit_from_tarball(*, project: Project, tarball: str, yaml_pat
     :return: Commit response object from API
     """
     yaml_path = yaml_path or project.get_yaml_path()
-    commit_obj = _get_pre_existing_commit(tarball)
+    commit_obj = _get_pre_existing_commit(tarball, project.id)
     if commit_obj:
         success(f"Ad-hoc code {commit_obj['identifier']} already uploaded")
     else:
@@ -67,7 +67,7 @@ def create_adhoc_commit_from_tarball(*, project: Project, tarball: str, yaml_pat
     return commit_obj
 
 
-def _get_pre_existing_commit(tarball: str) -> Optional[dict]:
+def _get_pre_existing_commit(tarball: str, project_id: str) -> Optional[dict]:
     try:
         # This is the same mechanism used by the server to
         # calculate the identifier for an ad-hoc tarball.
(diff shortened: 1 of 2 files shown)


valohai_cli/adhoc.py
```diff
@@ -59,23 +59,27 @@ def create_adhoc_commit_from_tarball(*, project: Project, tarball: str, yaml_pat
...
+    commit_obj = _get_pre_existing_commit(tarball, project.id)
     if commit_obj:
         success(f"Ad-hoc code {commit_obj['identifier']} already uploaded")
     else:
         commit_obj = _upload_commit_code(project=project, tarball=tarball, yaml_path=yaml_path, description=description)
     return commit_obj
 
 
-def _get_pre_existing_commit(tarball: str) -> Optional[dict]:
+def _get_pre_existing_commit(tarball: str, projectId: str) -> Optional[dict]:
```
akx, line 70: `snake_case`, please!

valohai_cli/utils/api_error_utils.py
```diff
@@ -35,6 +35,10 @@ def find_error(
     message: Optional[StringOrPattern] = None,
     matcher: Optional[Callable] = None,
 ) -> Optional[Any]:
+    if response_data is None:
+        # Error not found in response
+        return False
```
akx, line 40: How could this happen? Is there a code path that could pass `None` as `response_data`? 🤔

akx (changes requested):

valohai_cli/utils/api_error_utils.py
magdapoppins: Yeah, in the executions exec file:
```
        if find_error(error_json.get('environment'), code='does_not_exist'): <-- sometimes no 'environment' in that json
            yield (
                f'Run `{click.style("vh environments", bold=True)}` to see '
                f'the complete list of available environments.'
            )
```

magdapoppins: Perhaps the issue is that instead of picking out 'environment' from there we should just give it the whole response json, though?

akx: Ah, okay, that's fine. No, just looking for a `does_not_exist` in the full response will give false positives there.

akx (approved):

PR #220, opened 2022-10-14
Create dependabot.yml for version bumps

akx:
Refs #216, #217 – let's see if Dependabot can keep us up to date.

See https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuring-dependabot-version-updates

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,6 @@
+version: 2
+updates:
+  - package-ecosystem: pip
+    directory: "/"
+    schedule:
+      interval: daily


ruksi (approved):
could also just check for upgrade on the packages mentioned in the issue, but I see no harm in checking all dependencies 

ref https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#allow

akx:
@ruksi Yeah, I figured we can fine-tune the configuration later :)

ruksi:
haven't tried the Dependabot `pip` that sources from `setup.cfg` and without a lock file but let's see if it works :wink:

akx:
@ruksi Judging by the 5 new PRs that just appeared, it doth work...

akx:
Oh, right. 😅

PR #226, opened 2022-10-17
Split requirements files and update

akx:
Split files are required as we still support Python 3.6 in this library, and the ecosystem at large has moved on...


ruksi (approved):
split `dev` dependencies to `test` for unit tests (less strict definitions) and `lint` for code checks so we can upgrade our lint tools while tests can still be ran on EOL Python 3.6

lgtm

PR #234, opened 2022-10-24
Allow specifying --ca-file during login (or in settings)

### LINKED ISSUE #215: Enable a custom trusted root cert
Instead of not verifying SSL, verify it against a known trusted root cert.

akx:
Fixes #215


valohai_cli/commands/login.py
```diff
@@ -28,14 +28,16 @@
...
+@click.option('--ca-file', help='Path to bundle file or directory with trusted SSL certificates (this setting is persisted)')
 @yes_option
 def login(
     username: str,
     password: str,
     token: Optional[str],
     host: Optional[str],
     yes: bool,
-    verify_ssl: bool,
+    verify_ssl,
```
ruksi, line 39: `valohai_cli/commands/login.py:33: error: Function is missing a type annotation for one or more arguments  [no-untyped-def]`

ruksi (changes requested):
Mypy no hapy

ruksi (approved):
lgtm!

PR #242, opened 2022-11-04
🧑‍💻🔨 Improve developer experience

ruksi:
Slight improvements to `make` targets and README developer guide.

I first tried running the development tooling with Python 3.7 (`mypy` fails to install the listed version) and then 3.8 (type checks give false positives) and jumped to Python 3.10 (based on version used by the CI) which works fine~~

Noted that down in the README.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -2,6 +2,14 @@
 default:
 	@echo "check README for usage"
 
+.PHONY: dev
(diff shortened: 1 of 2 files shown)


magdapoppins (approved):
Awesome! ✨

PR #247, opened 2022-11-09
♻️✅ Conflate how git is managed in tests

ruksi:
Unit test git usage was done a few different ways in a few different places and this streamlines that.

Also adds a few smoke tests to git utils.

Extracted from another thing I was testing related to custom YAML paths in local mode.


tests/test_packaging.py
```diff
@@ -51,16 +52,17 @@ def get_expected_filenames(original_set: Set[str], *, with_gitignore, with_vhign
 @pytest.mark.parametrize('with_commit', (False, True))
 @pytest.mark.parametrize('with_vhignore', (False, True))
 def test_package_git(tmpdir, with_commit, with_vhignore):
```
ruksi, line 54: `tmpdir` should be replaced with the more recent `tmp_path` but decided to leave that out of this PR

magdapoppins (approved):
lgtm!

PR #272, opened 2023-01-18
Lint with pre-commit + ruff

akx:
Hot on the heels of https://github.com/valohai/valohai-yaml/pull/84 :)


ruksi (approved):
`ruff` it up! 🐶

PR #273, opened 2023-02-07
Fix exec remote debug API call correct format

DrazenDodik:
Small fix to move `debug_port ` and `debug_key` under `remote_debug` so the API call gets sent in the expected format


akx (approved):
Good catch!

PR #278, opened 2023-03-10
Add missing slash into authtoken url

tokkoro:
The trailing slash is optional when using custom host, but token generation URL format expects it.

Fixed 
image


ruksi (approved):
slash it up!

PR #281, opened 2023-08-15
Honour --yaml flag to parse configuration in vh pipeline run --adhoc

hylje:
Resolves https://github.com/valohai/roi/issues/5071. 

During `vh pipeline run <pipeline> --adhoc`, pipeline steps were resolved on CLI end without honouring the `--yaml` flag for custom configuration. Project default configuration file's steps were used always. We shall now honour the flag. 

This behaviour has an existing test case and the tested code raised an exception as it did not find the default configuration file. However, the default test runner catches exceptions and the test case did not check that the command finished with a successful exit code, nor expected enough output to determine success. These oversights allowed the test to succeed. We will now reraise any exceptions raised during command tests, and pipeline tests will positively assert success up to the API response.

Valohai backend works as expected with CLI initialising its pipeline creation API call using custom configuration.

### DIFF
--- a/tests/commands/pipeline/test_run.py
+++ b/tests/commands/pipeline/test_run.py
@@ -13,17 +13,20 @@ def test_pipeline_run_success(runner, logged_in_and_linked):
     args = ['training']
     with RunAPIMock(PROJECT_DATA['id']):
         output = runner.invoke(run, args).output
+
     assert 'Success' in output
+    assert 'Pipeline =21 queued' in output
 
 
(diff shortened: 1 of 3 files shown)


ruksi (approved):
> We will now reraise any exceptions raised during command tests, and pipeline tests will positively assert success up to the API response.

👍 💯 

sounds and looks reasonable

could've added the `PedanticCliRunner` in a separate commit to separate if from the actual solution for the issue edits but good enough!

PR #282, opened 2023-08-15
Update lint & test instructions

teroyks:
Old Make recipies have been replaced with pre-commit.
Bring the README up to date with the current configuration.

(Current actions copied from the yaml repo README.)

### DIFF
--- a/README.md
+++ b/README.md
@@ -89,7 +89,15 @@ vh --help
 To run lints, type checks and tests:
 
 ```bash
-make lint mypy test
+# setup development dependencies
+pip install -e . -r requirements-test.txt pre-commit
+pre-commit install
+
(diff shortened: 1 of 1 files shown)


ruksi (approved):
a good DX update!

I would've done it by first updating the `Makefile` as the is already the `dev` dependency install target:

```Makefile
.PHONY: dev
dev:
	pip install -U pip
	pip install -r requirements-test.txt
	pip install pre-commit
	pip install -e .
	pre-commit install
```

then add a `lint` target:

```Makefile
.PHONY: lint
lint:
	pre-commit run --all-files
```

so it would be `make lint test` as pre-commit will run the `mypy`

but we can change that later if we feel like it 👍

PR #283, opened 2023-08-16
Create a make recipe to run lint

teroyks:
Linting still done with pre-commit under the hood. Add pre-commit installation to `make dev` so that a separate installation step is not needed.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -6,7 +6,13 @@ default:
 dev:
 	pip install -U pip
 	pip install -r requirements-test.txt
+	pip install pre-commit
 	pip install -e .
+	pre-commit install
+
+.PHONY: lint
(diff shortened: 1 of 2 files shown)


teroyks:
@ruksi re: [previous PR](https://github.com/valohai/valohai-cli/pull/282#pullrequestreview-1579819995) – I would prefer `make` recipes as well – wasn’t sure if the goal is to move away from `make`, but here it is anyway. 😄

ruksi (approved):
I'm not sure if we have any official stand regarding Make but I do like it more than having to copy-paste commands from the README

`git clone` to `make <tab tab tab>` and so on :stuck_out_tongue: 

looking excellent!

PR #288, opened 2023-09-11
Update pre-commit tools, add more Ruff rules, fix 'em

akx:
The usual cleanup.


ruksi (approved):
up 'em, ruff 'em, stick 'em in a stew

Issue #279, opened 2023-04-12
Add description of valohai-cli commands


ruslands:
`vh --help` doesn't show details of the commands.
`vh execution --help ` --ad-hoc flag is missing in the options.

![Screenshot 2023-04-12 at 08 25 52](https://user-images.githubusercontent.com/16615834/231359258-0935c15a-adf2-497a-a980-e4f118eb586a.png)

![Screenshot 2023-04-12 at 08 29 25](https://user-images.githubusercontent.com/16615834/231359560-94096d1b-00b9-470c-8d8b-84caed184f37.png)

PR #290, opened 2023-09-11
Ensure help texts are shown (and all commands have them)

akx:
The clever idea to use an ExceptionGroup (the other option having been [pytest-check](https://pypi.org/project/pytest-check/)) was courtesy of tpr at koodiklinikka.

Also, turns out the original #279 bug was just a matter of newline padding in the help string; `get_short_help_str` had been added to Click only some time after I'd written the original plugin CLI code, so no wonder we hadn't used it: https://github.com/pallets/click/pull/869

Click 8 is required for `to_info_dict`: https://github.com/pallets/click/pull/1623

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -13,7 +13,7 @@ authors = [
     { name = "Valohai", email = "hait@valohai.com" },
 ]
 dependencies = [
-    "click>=7.0",
+    "click>=8.0",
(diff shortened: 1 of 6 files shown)


teroyks (approved):
Beautiful dev-quality-of-life improvement!

PR #292, opened 2023-10-02
Add `fetch commits` functionality in case of unknown commit

### LINKED ISSUE #254: Incorrect error message when commit is not yet fetched
Valohai-cli should ask if user want to fetch after Roi fails to find the commit. 

### User story:
When trying to run execution, got:

(research) MacBook-Pro-9:language\_modeling\_research $ vh execution run train  
😿 Warning: Commit e2780cc5ef7c54c4b56fb69cb2ca21cd06f81791 is not known for the project. Have you pushed it?

...  
but I had actually pushed the commit. Actually I had to click "Fetch repository" on the app page. I wonder why it cannot automatically fetch the repo? gz#1194

memona008:
Resolves #254 

#Why does this pull request exists?
When there is new commits in git repository, cli wont fetch the latest commits and abort by declaring the commit unknown.

##Changes made in PR:
- Fetch latest details when a commit is unknown
- Modified tests according to the case

### DIFF
--- a/valohai_cli/api.py
+++ b/valohai_cli/api.py
@@ -72,10 +72,10 @@ class APISession(requests.Session):
         try:
             resp = super().request(method, url, **kwargs)
         except requests.ConnectionError as ce:
-            host = urlparse(ce.request.url).netloc
+            host = urlparse(ce.request.url).netloc if ce.request else self.base_netloc
             if 'Connection refused' in str(ce):
                 raise CLIException(
-                    f'Unable to connect to {host} (connection refused); try again soon.'
+                    f'Unable to connect to {host!r} (connection refused); try again soon.'
                 ) from ce
             raise
 
--- a/tests/test_utils_commits.py
+++ b/tests/test_utils_commits.py
@@ -78,14 +78,29 @@ class TestResolveCommit:
         out, err = capsys.readouterr()
         assert 'which is ambiguous with' in err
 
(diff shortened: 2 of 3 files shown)


valohai_cli/api.py
```diff
@@ -72,10 +72,10 @@ def request(self, method: str, url: str, **kwargs: Any) -> Response:  # type: ig
         try:
             resp = super().request(method, url, **kwargs)
         except requests.ConnectionError as ce:
-            host = urlparse(ce.request.url).netloc
+            host = urlparse(ce.request.url).netloc if ce.request else self.base_netloc
             if 'Connection refused' in str(ce):
                 raise CLIException(
-                    f'Unable to connect to {host} (connection refused); try again soon.'
+                    f'Unable to connect to {host!r} (connection refused); try again soon.'
```
ruksi, line 78: an error is shown here because `host` can be either a `str` or `bytes`, depending on which was given as the parameter to the `urlparse()` function

in this context we just work with `str` so should always be that, but if it somehow would be `bytes`, then the `{host}` would translate to `b'abc'`, not `abc` (no quotes)

with this change
+ `bytes` will still be `b'abc'`, like before
+ `str` becomes from `abc` to `abc` (the representation)

so now the error message host name is effectively always surrounded with quotes

so this just to silence the (rightful) lint warning about a potential issue

ruksi (approved):
lgtm :+1:

PR #298, opened 2023-11-29
Kubernetes step parsing and command line options

hylje:
Included in this PR:
===

1. Parse Kubernetes `resource` structure from YAML when using `vh execution run`, if provided.
2. Parse Kubernetes resource claims directly as `vh execution run` command line arguments

`vh pipeline run` does not have execution runtime option override support for now.

### DIFF
--- a/tests/commands/execution/test_run.py
+++ b/tests/commands/execution/test_run.py
@@ -2,7 +2,7 @@ import pytest
 import yaml
 
 from tests.commands.run_test_utils import ALTERNATIVE_YAML, RunAPIMock, RunTestSetup
-from tests.fixture_data import CONFIG_YAML, PROJECT_DATA
+from tests.fixture_data import CONFIG_YAML, KUBE_RESOURCE_YAML, PROJECT_DATA
 from valohai_cli.commands.execution.run import run
 from valohai_cli.ctx import get_project
 from valohai_cli.models.project import Project
@@ -13,6 +13,15 @@ adhoc_mark = pytest.mark.parametrize('adhoc', (False, True), ids=('regular', 'ad
 @pytest.fixture(params=['regular', 'adhoc'], ids=('regular', 'adhoc'))
 def run_test_setup(request, logged_in_and_linked, monkeypatch):
     return RunTestSetup(monkeypatch=monkeypatch, adhoc=(request.param == 'adhoc'))
+
+
+@pytest.fixture(params=['regular', 'adhoc'], ids=('regular', 'adhoc'))
+def run_test_setup_kube(request, logged_in_and_linked, monkeypatch):
+    return RunTestSetup(
+        monkeypatch=monkeypatch,
+        adhoc=(request.param == 'adhoc'),
+        config_yaml=KUBE_RESOURCE_YAML
+    )
 
 
 @pytest.fixture()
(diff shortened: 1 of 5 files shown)


tests/commands/execution/test_run.py
```diff
@@ -252,3 +261,133 @@ def test_remote(run_test_setup, tmpdir):
...
+    run_test_setup.args.append("--k8s-cpu-min=1")
+    run_test_setup.args.append("--k8s-memory-min=2")
+    run_test_setup.args.append("--k8s-cpu-max=3")
+    run_test_setup.args.append("--k8s-memory-max=4")
+    run_test_setup.args.append("--k8s-device=nvidia.com/gpu=1")
+    run_test_setup.run()
+
+    assert run_test_setup.run_api_mock.last_create_execution_payload["runtime_config"] == {
+        "kubernetes": {
+            'containers': {
```
ruksi, line 276: mixing of the `"` and `'` hurts my souls a bit but doesn't really matter :ghost:

tests/commands/execution/test_run.py
```diff
@@ -252,3 +261,133 @@ def test_remote(run_test_setup, tmpdir):
...
+                'workload': {
+                    'resources': {
+                        'requests': {
+                            'cpu': 1.0,
+                        }
+                    }
+                }
+            }
+        }
+    }
```
ruksi, line 313: the rest are not sent in the payload as we except that the API fills them from the YAML :+1:

tests/commands/execution/test_run.py
```diff
@@ -252,3 +261,133 @@ def test_remote(run_test_setup, tmpdir):
...
+                        'limits': {
+                            'cpu': 2.0,
+                            'memory': 4,
+                            'devices': {}
+                        }
+                    }
+                }
+            }
+        }
+    }
```
ruksi, line 393: not sure I understand this, why are the non-gpu resources sent in this case?

ruksi (changes requested):
lgtm, but a _question_

tests/commands/execution/test_run.py
hylje: Resources from YAML are parsed and sent if available (see the `run_test_setup_kube` fixture), and command line arguments can override them further. `--k8s-device-none` overrides just the devices dict into an explicit empty dict, while other resources are not overridden.

In the above partial case, the YAML doesn't define any resources so we won't send them either

ruksi: but if we send YAML defined resources, wouldn't that mix up e.g. future copying? (we don't know know which of the passed parameters are API overrides vs. YAML overrides)

assuming that API doesn't compare passed resources to the YAML it knows 🤔

ruksi: but I guess it's fine if it work like the wen UI, could double check it

hylje: I don't think executions keep track of where the resulting configuration comes from, it's all put together and saved for later use. Copying doesn't refresh changes from yaml Steps, whether or not they were also supplied through the API or not.

I think the principle that front-end resolves the YAML itself and submits from it is alright, we already do it in CLI for parameters, inputs and environment variables.

I'm not against just leaving it for the API backend to work out, but I think that's its own project.

ruksi: > I think the principle that front-end resolves the YAML itself and submits from it is alright

yeah, I also feel that it's all-OK :+1: 

> I'm not against just leaving it for the API backend to work out, but I think that's its own project

yeah, definitely a separate thing if we feel that it's needed

ruksi (approved):
is good

PR #304, opened 2024-01-24
Update pre-commit hooks to latest versions

teroyks:
Running some pretty old versions.
Autoupdate to current.


hylje (approved):
🧹 gotta keep the factory floor clean

PR #305, opened 2024-01-24
Add command line parameter for runtime config preset

### LINKED ISSUE #303: Add support for setting a runtime config preset in CLI option
Add an option (`--k8s-preset PRESET_UUID`) for `execution run` that will add the preset to the API call.

teroyks:
Resolves #303 

Allow setting a preset ID with the `--k8s-preset` command line option.

Note: preset slug support will be automatic once that has been added to the API – no functional changes required (though the help text should be updated).

_Note:_ The first commit contains all the functional changes in this PR – the rest are just refactoring and tidying up the code a bit. Especially the last one is things that are more a matter of taste (and can be reverted easily).

### DIFF
--- a/tests/commands/execution/test_run.py
+++ b/tests/commands/execution/test_run.py
@@ -391,3 +391,18 @@ def test_kube_step_override_device_empty(run_test_setup_kube):
             }
         }
     }
+
+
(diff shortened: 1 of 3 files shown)


hylje (approved):
lgtm, tests are good too

PR #310, opened 2024-05-24
Smoother token login flow

### LINKED ISSUE #307: Allow logging in via browser
Currently, the CLI allows logging using either a username/password or via a token. However, the username/password option doesn't work when two-factor authentication is enabled or when SSO is the preferred sign-in mode. The token works in the latter cases, but generating one is cumbersome and difficult to discover. 

For a more convenient approach that doesn't require memorizing passwords, copy/pasting strings, or generating tokens, it would be good to allow logging in by opening the browser.

akx:
Instead of telling the user to run a command to log in with the token they generate, we ask for it interactively.

Fixes #307.

In lieu of a screenshot...
```
$ vh login --host https://valohai.local/

If you don't yet have an account, please create one at https://valohai.local/ first.

https://valohai.local/ - Username: 2faman
2faman on https://valohai.local/ - Password: [...]
Retrieving API token from https://valohai.local/...
======================================================================================
Oops!
The error code "has_2fa" indicates username + password authentication is not possible.
Use a login token instead:
 1. Log in on https://valohai.local/
 2. Visit https://valohai.local/auth/tokens/ to generate an authentication token
 3. Once you have an authentication token, paste it below and we'll try it out.
======================================================================================
Enter generated token: [...]
Verifying API token on https://valohai.local/...
😄  Success! Logged in. Hey 2faman!
```

### DIFF
--- a/valohai_cli/commands/login.py
+++ b/valohai_cli/commands/login.py
@@ -17,8 +17,7 @@ The error code "{code}" indicates username + password authentication is not poss
 Use a login token instead:
  1. Log in on {host}
  2. Visit {token_url} to generate an authentication token
- 3. Once you have an authentication token, log in with:
-    {command}
+ 3. Once you have an authentication token, paste it below and we'll try it out.
 '''.strip()
 
 
@@ -78,10 +77,7 @@ def login(
             verify_ssl=verify_ssl,
         )
 
-    click.echo(f'Verifying API token on {host}...')
-
-    with APISession(host, token, verify_ssl=verify_ssl) as sess:
-        user_data = sess.get('/api/v0/users/me/').json()
+    user_data = verify_token(host=host, token=token, verify_ssl=verify_ssl)
     settings.persistence.update(
         host=host,
         user=user_data,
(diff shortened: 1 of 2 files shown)


tarekoraby:
I assume that `valohai.local` will be replaced by `app.valohai.com` in production. Is that correct? If so, should we care about on-prem installations (i.e., should we tell users that they might want to replace `app.valohai.com` with their own org url)?

valohai_cli/commands/login.py
```diff
@@ -117,20 +119,36 @@ def do_user_pass_login(
...
+    verify_ssl: Union[bool, str] = True,
+    login_error_code: Optional[str] = None,
+) -> str:
+    banner(TOKEN_LOGIN_HELP.format(
+        code=login_error_code or "(unknown)",
+        host=click.style(host, bold=True),
+        token_url=click.style(urljoin(host, '/auth/tokens/'), bold=True),
+    ))
+    while True:
+        token = click.prompt('Enter generated token', hide_input=True)
```
dangquangdon, line 143: I understand that we want to hide the input for safety reason, however, I wonder, from user's perspective, should we allow them to see what they paste to the terminal, at least they can double check if they paste the right thing 🤔

dangquangdon:
> I assume that `valohai.local` will be replaced by `app.valohai.com` in production. Is that correct? If so, should we care about on-prem installations (i.e., should we tell users that they might want to replace `app.valohai.com` with their own org url)?

By default the CLI is pointing to `app.valohai.com`, but user can use `--valohai-host` to override it with whatever host they use. Iirc, it was also mentioned in the academy 🤔

valohai_cli/commands/login.py
akx: If they paste the wrong thing, the check will fail, and the loop will continue.

akx:
> Should we care about on-prem installations?

Yes, the messaging will use whichever host the user specifies (updated the "screenshot" to reflect that I actually pointed the CLI to my local install).

dangquangdon (approved):

PR #311, opened 2024-06-11
Support list parameters for pipelines

hylje:
Extends pipeline command line parameter parsing to allow overriding the default pipeline parameters with a list parameter

Upgrades `valohai-yaml` dependency to 0.40.0 containing list-type pipeline parameter support as well as override support in PipelineConverter which we will pass any command line overrides to


memona008 (approved):
LGTM

PR #313, opened 2024-08-29
option to disable git packaging for adhoc execution

### LINKED ISSUE #122: Submodules in ad-hoc runs
`git ls-files` does not delve into submodule directories, so when using Git to package files for ad-hoc runs, those files don't get uploaded.

`git ls-files --recurse-submodules` is only supported with `--cached` at present.

A workaround is to hack the packager not to use git for packaging.

Options:

* Expose the `allow_git` flag (e.g. an envvar/global click flag `--no-git`)
* If we'd otherwise use `git ls-files`, detect if submodules are in use and don't use `git ls-files` if they are. (e.g. if the output of `git submodule` is longer than zero).

Kenneth-Schroeder:
should resolve #122

### DIFF
--- a/valohai_cli/commands/pipeline/run/run.py
+++ b/valohai_cli/commands/pipeline/run/run.py
@@ -21,7 +21,7 @@ from valohai_cli.utils.commits import create_or_resolve_commit
 @click.option('--commit', '-c', default=None, metavar='SHA', help='The commit to use. Defaults to the current HEAD.')
 @click.option('--title', '-c', default=None, help='The optional title of the pipeline run.')
 @click.option('--adhoc', '-a', is_flag=True, help='Upload the current state of the working directory, then run it as an ad-hoc execution.')
-@click.option('--disable_git_packaging', '-g', is_flag=True, default=False, help='When using ad-hoc execution, whether to disable git for packaging directory contents.')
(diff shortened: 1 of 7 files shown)


valohai_cli/commands/pipeline/run/run.py
```diff
@@ -51,7 +53,7 @@ def run(
     if yaml and not adhoc:
         raise click.UsageError('--yaml is only valid with --adhoc')
 
-    commit = create_or_resolve_commit(project, commit=commit, adhoc=adhoc, yaml_path=yaml)
+    commit = create_or_resolve_commit(project, commit=commit, adhoc=adhoc, yaml_path=yaml, allow_git_packaging=not disable_git_packaging)
```
akx, line 56:
```suggestion
    commit = create_or_resolve_commit(project, commit=commit, adhoc=adhoc, yaml_path=yaml, allow_git_packaging=git_packaging)
```

akx (changes requested):
Thanks!

I think a positive argument that's set to `true` by default, and can be disabled with a negative flag, would be more intuitive.

NB: I didn't test the suggestions locally, so they might be wrong :)

Kenneth-Schroeder:
Thank you, good feedback, I am not familiar with click :)

teroyks (approved):
LGTM, and the test agrees.

PR #314, opened 2024-08-31
Add SSH command for executions 

### LINKED ISSUE #312: Add remote-debug command
User will be able to remotely debug the execution without doing any manual steps 
Command will look like 
vh exec ssh <exec-number>

memona008:
Resolves #312 
User will be able to remotely debug the execution 
1. He can either provide address:port & private key by himself or run the command for auto collection of the details
2. If the key is generated in CLI while starting the execution it can be found by command and prompted to user as option 
3. But if the execution is started by UI or key is generated somewhere user HAS to provide that key as command argument `--private-key-file `

[ssh.webm](https://github.com/user-attachments/assets/ec4c3d48-5bb8-4f48-9a88-a2ff7f9586bf)

### DIFF
--- a/tests/commands/execution/test_ssh.py
+++ b/tests/commands/execution/test_ssh.py
@@ -1,0 +1,80 @@
+import shutil
+import subprocess
+import time
+
+from tests.commands.execution.utils import get_execution_data_mock, no_sleep
+from tests.fixture_data import EXECUTION_DATA, STATUS_EVENT_RESPONSE_DATA
+from valohai_cli.commands.execution.ssh import ssh
+
+
+def test_ssh_in_completed_execution(runner, logged_in_and_linked, monkeypatch):
+    with get_execution_data_mock():
+        counter = EXECUTION_DATA["counter"]
+        result = runner.invoke(ssh, [str(counter)], catch_exceptions=False)
+        assert f"Error: Execution #{counter} is complete. Cannot SSH into it.\n" in result.output
+        assert result.exit_code == 1
+
+
+def test_ssh_in_queued_execution(runner, logged_in_and_linked, monkeypatch):
+    counter = EXECUTION_DATA["counter"]
+    monkeypatch.setitem(EXECUTION_DATA, "status", "queued")
+    monkeypatch.setattr(time, "sleep", no_sleep)
+    with get_execution_data_mock():
+        result = runner.invoke(ssh, [str(counter)], catch_exceptions=False)
+        assert f"Execution #{counter} is queued. Waiting for it to start...\n" in result.output
+        assert result.exit_code == 1
+
+
+def test_ssh_with_no_ssh_details_present(runner, logged_in_and_linked, monkeypatch):
+    counter = EXECUTION_DATA["counter"]
+    monkeypatch.setitem(EXECUTION_DATA, "status", "started")
+    monkeypatch.setattr(time, "sleep", lambda x: None)
+    with get_execution_data_mock() as m:
+        m.get(
+            f"https://app.valohai.com/api/v0/executions/{EXECUTION_DATA['id']}/status-events/",
+            json={"status_events": []},
+        )
+        result = runner.invoke(ssh, [str(counter)], catch_exceptions=False)
+        output = result.output
+        assert "1/5 Retrying: No SSH details found...\n" in output
+        assert "2/5 Retrying: No SSH details found...\n" in output
+        assert "3/5 Retrying: No SSH details found...\n" in output
+        assert "4/5 Retrying: No SSH details found...\n" in output
+        assert "5/5 Retrying: No SSH details found...\n" in output
+
+    assert result.exit_code == 1
+
+
+def test_ssh(runner, logged_in_and_linked, monkeypatch, tmp_path):
+    counter = EXECUTION_DATA["counter"]
+    monkeypatch.setitem(EXECUTION_DATA, "status", "started")
+
+    def mock_prompt():
+        return tmp_path
+
+    monkeypatch.setattr(
+        "valohai_cli.commands.execution.ssh.select_private_key_from_possible_directories",
+        mock_prompt,
+    )
+
+    with get_execution_data_mock() as m:
+        m.get(
+            f"https://app.valohai.com/api/v0/executions/{EXECUTION_DATA['id']}/status-events/",
+            json=STATUS_EVENT_RESPONSE_DATA,
+        )
+        result = runner.invoke(ssh, [str(counter)], catch_exceptions=False)
+        output = result.output
+        assert "SSH address is 127.0.0.1:2222" in output
+
+        def mock_subprocess_run(*args, **kwargs):
+            print(args[0])
+            return subprocess.CompletedProcess(args=args, returncode=0)
+
+        monkeypatch.setattr(subprocess, "run", mock_subprocess_run)
+
+        result = result.runner.invoke(ssh, [str(counter)], input="1", catch_exceptions=False)
+        assert (
+            f"['{shutil.which('ssh')}', '-i', PosixPath('{tmp_path}'), 'root@127.0.0.1', '-p', '2222', '-t', '/bin/bash']"
+            in result.output
+        )
+        assert result.exit_code == 0
--- a/tests/fixture_data.py
+++ b/tests/fixture_data.py
@@ -330,7 +330,6 @@ CONFIG_YAML = """
         path: model.h5
 """
 
-
 PIPELINE_YAML = (
     CONFIG_YAML
     + """
@@ -665,3 +664,19 @@ KUBE_RESOURCE_YAML = """
         description: Model output file from TensorFlow
         path: model.h5
 """
+
+STATUS_EVENT_RESPONSE_DATA = {
+    "total": 2,
+    "status_events": [
+        {
+            "stream": "status",
+            "message": '::ssh::{"port": 2222, "address": "127.0.0.1"}',
+            "time": "2024-09-04T12:16:20.722000",
+        },
+        {
+            "stream": "status",
+            "message": "   $ ssh -i <path-to-private-key> 127.0.0.1 -p 2222 -t /bin/bash",
+            "time": "2024-09-04T12:16:20.723000",
+        },
+    ],
+}
--- a/valohai_cli/commands/execution/run/dynamic_run_command.py
+++ b/valohai_cli/commands/execution/run/dynamic_run_command.py
@@ -18,6 +18,7 @@ from valohai_cli.utils import humanize_identifier, sanitize_option_name
 from valohai_cli.utils.file_input import read_data_file
 from valohai_cli.utils.friendly_option_parser import FriendlyOptionParser
 
+from ..ssh import ssh
 from .excs import ExecutionCreationAPIError
 
 
@@ -59,6 +60,7 @@ class RunCommand(click.Command):
         tags: Optional[Sequence[str]] = None,
         runtime_config: Optional[dict] = None,
         runtime_config_preset: Optional[str] = None,
+        ssh: bool = False,
     ) -> None:
         """
         Initialize the dynamic run command.
@@ -75,6 +77,7 @@ class RunCommand(click.Command):
         :param download_directory: Where to (if somewhere) to download execution outputs (sync mode)
         :param runtime_config: Runtime config dict
         :param runtime_config_preset: Runtime config preset identifier (UUID)
+        :param ssh: Whether to chain to `exec ssh` afterward
         """
         assert isinstance(step, Step)
         self.project = project
@@ -90,6 +93,7 @@ class RunCommand(click.Command):
         self.tags = list(tags or [])
         self.runtime_config = dict(runtime_config or {})
         self.runtime_config_preset = runtime_config_preset
+        self.ssh = ssh
         super().__init__(
             name=sanitize_option_name(step.name.lower()),
             callback=self.execute,
@@ -200,6 +204,12 @@ class RunCommand(click.Command):
             import webbrowser
 
             webbrowser.open(resp["urls"]["display"])
+
+        if self.ssh:
+            try:
+                ctx.invoke(ssh, counter=resp["counter"])
+            except Exception as e:
+                warn(f"Failed to open SSH connection: {e}")
 
         if self.watch:
             from valohai_cli.commands.execution.watch import watch
--- a/valohai_cli/commands/execution/run/frontend_command.py
+++ b/valohai_cli/commands/execution/run/frontend_command.py
@@ -110,6 +110,11 @@ EMPTY_DICT_PLACEHOLDER = object()
 @click.option("--debug-port", type=int)
 @click.option("--debug-key-file", type=click.Path(file_okay=True, readable=True, writable=False))
 @click.option(
+    "--ssh",
+    is_flag=True,
+    help="Start ssh remote connection for debugging the execution after it starts.",
+)
+@click.option(
     "--autorestart/--no-autorestart",
     help="Enable Automatic Restart on Spot Instance Interruption",
 )
@@ -184,6 +189,7 @@ def run(
     k8s_devices: List[str],
     k8s_device_none: bool,
     k8s_preset: Optional[str],
+    ssh: bool = False,
 ) -> Any:
     """
     Start an execution of a step.
@@ -197,8 +203,8 @@ def run(
     project = get_project(require=True)
     project.refresh_details()
 
-    if download_directory and watch:
-        raise click.UsageError("Combining --sync and --watch not supported yet.")
+    if sum([watch, download_directory is not None, ssh]) > 1:
+        raise click.UsageError("Only one of --watch, --sync or --ssh can be set.")
 
     if not commit and project.is_remote:
         # For remote projects, we need to resolve early.
@@ -239,6 +245,7 @@ def run(
             "debug_port": debug_port,
             "debug_key": key,
         }
+
     if autorestart:
         runtime_config["autorestart"] = autorestart
 
@@ -291,6 +298,7 @@ def run(
         tags=tags,
         runtime_config=runtime_config,
         runtime_config_preset=k8s_preset,
+        ssh=ssh,
     )
     with rc.make_context(rc.name, list(args), parent=ctx) as child_ctx:
         return rc.invoke(child_ctx)
--- a/valohai_cli/commands/execution/ssh.py
+++ b/valohai_cli/commands/execution/ssh.py
@@ -1,0 +1,46 @@
+import click
+
+from valohai_cli.exceptions import CLIException
+from valohai_cli.utils.cli_utils import counter_argument
+from valohai_cli.utils.ssh import (
+    get_ssh_details_with_retry,
+    make_ssh_connection,
+    select_private_key_from_possible_directories,
+)
+
+
+@click.command()
+@counter_argument
+@click.option(
+    "--private-key-file",
+    default=None,
+    type=click.Path(file_okay=True, exists=True),
+    help="Private SSH key to use for the connection.",
+)
+@click.option(
+    "--address",
+    default=None,
+    help='Address of the container in "ip:port" format. If not provided, '
+    "the address from the execution will be used.",
+)
+def ssh(counter: int, private_key_file: str, address: str) -> None:
+    """
+    Make SSH Connection to the execution container.
+    """
+    if address:
+        try:
+            ip_address, _, port_str = address.partition(":")
+            if not ip_address or not port_str:
+                raise CLIException("Address must be in 'ip:port' format.")
+            port = int(port_str)
+            if port <= 1023:
+                raise CLIException("Port must be above 1023")
+        except ValueError as e:
+            raise CLIException(f"Invalid address format: {e}")
+    else:
+        ip_address, port = get_ssh_details_with_retry(counter)
+
+    click.echo(f"SSH address is {ip_address}:{port}")
+    if not private_key_file:
+        private_key_file = select_private_key_from_possible_directories()
+    make_ssh_connection(ip_address, port, private_key_file)
--- a/valohai_cli/utils/__init__.py
+++ b/valohai_cli/utils/__init__.py
@@ -3,9 +3,10 @@ import os
 import random
 import re
 import string
+import time
 import unicodedata
 import webbrowser
-from typing import Any, Callable, Dict, Iterable, Iterator, Tuple, Union
+from typing import Any, Callable, Dict, Iterable, Iterator, Tuple, Type, TypeVar, Union
 
 import click
 
@@ -154,3 +155,24 @@ def parse_environment_variable_strings(
 
 def compact_dict(dct: dict) -> dict:
     return {key: value for (key, value) in dct.items() if key and value}
+
+
+T = TypeVar("T")
+
+
+def call_with_retry(
+    func: Callable[[], T],
+    retries: int = 3,
+    delay_range: Tuple[int, int] = (1, 5),
+    retry_on_exception_classes: Tuple[Type[Exception], ...] = (Exception,),
+) -> T:
+    for attempt in range(retries):
+        try:
+            return func()
+        except retry_on_exception_classes as e:
+            click.echo(f"{attempt + 1}/{retries} Retrying: {e}...")
+            if attempt + 1 == retries:
+                raise RuntimeError(f"Failed after {retries} attempts") from e
+            time.sleep(random.uniform(*delay_range))
+
+    raise RuntimeError("Shouldn't get here!")
--- a/valohai_cli/utils/ssh.py
+++ b/valohai_cli/utils/ssh.py
@@ -1,0 +1,144 @@
+import json
+import os
+import pathlib
+import shutil
+import subprocess
+import time
+from dataclasses import dataclass
+from itertools import count
+from typing import Generator, Iterable, Iterator, Optional, Tuple
+
+import click
+
+from valohai_cli.api import request
+from valohai_cli.ctx import get_project
+from valohai_cli.exceptions import CLIException
+from valohai_cli.settings import get_settings_file_name
+from valohai_cli.utils import call_with_retry
+from valohai_cli.utils.cli_utils import prompt_from_list
+
+
+@dataclass(frozen=True)
+class Keypair:
+    public_path: pathlib.Path
+    private_path: pathlib.Path
+
+
+def make_ssh_connection(address: str, port: int, private_key_path: str) -> None:
+    ssh_path = shutil.which("ssh")
+    if ssh_path is None:
+        raise FileNotFoundError(
+            "SSH command not found. Ensure SSH is installed and available in your PATH.",
+        )
+    ssh_command = [
+        ssh_path,
+        "-i",
+        private_key_path,
+        f"root@{address}",
+        "-p",
+        str(port),
+        "-t",
+        "/bin/bash",
+    ]
+    subprocess.run(ssh_command)
+
+
+def find_ssh_details(events: Iterable) -> Optional[Tuple[str, int]]:
+    prefix = "::ssh::"
+    for event in events:
+        if event["message"].startswith(prefix):
+            data = json.loads(event["message"][len(prefix) :])
+            return data["address"], int(data["port"])
+    return None
+
+
+def get_ssh_key_pairs_from_dir(dir: str) -> Iterator[Keypair]:
+    folder_path = pathlib.Path(dir)
+    for pth in folder_path.glob("*.pub"):
+        if not pth.is_file():
+            continue
+        try:
+            with pth.open("rb") as f:
+                first_line = f.readline().strip()
+                if first_line.startswith((b"ssh-rsa", b"ecdsa-sha2-", b"ssh-ed25519")):
+                    private_key_paths = [pth.with_suffix(""), pth.with_suffix(".pem")]
+                    for private_key_path in private_key_paths:
+                        if private_key_path.is_file():
+                            yield Keypair(public_path=pth, private_path=private_key_path)
+                            break
+        except OSError:  # e.g. unreadable key
+            pass
+
+
+def select_private_key_from_possible_directories() -> str:
+    default_ssh_dir = os.path.expanduser("~/.ssh")
+    cli_config_dir = get_settings_file_name("")
+    all_keypairs: list[Keypair] = []
+    all_dirs = [cli_config_dir, default_ssh_dir]
+    for dir in all_dirs:
+        if not os.path.isdir(dir):
+            continue
+        all_keypairs.extend(get_ssh_key_pairs_from_dir(dir))
+
+    if not all_keypairs:
+        raise click.UsageError(
+            f"No SSH key pairs found in {', '.join(all_dirs)}. Please try to generate new SSH key pair",
+        )
+
+    keypair_options = [
+        {
+            "name": keypair.private_path,
+            "description": f"Public: {keypair.public_path}",
+        }
+        for keypair in all_keypairs
+    ]
+
+    selected_key = prompt_from_list(prompt="Select the SSH key pair you want to use", options=keypair_options)
+    return str(selected_key["name"])
+
+
+def fetch_status_events(execution: dict, first_n: Optional[int] = None) -> Generator:
+    params = {}
+    if first_n is not None:
+        params["limit"] = first_n
+    events_response = request("get", f"{execution['url']}status-events/", params=params).json()
+    yield from events_response.get("status_events", [])
+
+
+def get_ssh_details_from_execution(execution: dict) -> Tuple[str, int]:
+    events_response = fetch_status_events(execution, first_n=100)
+    ssh_details = find_ssh_details(events_response)
+    if not ssh_details:
+        raise click.UsageError("No SSH details found")
+    return ssh_details
+
+
+def get_ssh_details_with_retry(
+    counter: int,
+) -> Tuple[str, int]:
+    """
+    Fetch SSH details for a given counter, retrying until the execution has started.
+
+    :param counter: The execution counter.
+    :return: A tuple of (IP address, port) for the SSH connection.
+    """
+    project = get_project(require=True)
+    assert project
+    for attempt in count():
+        execution = project.get_execution_from_counter(
+            counter=counter,
+            params={"exclude": "metadata, events"},
+        )
+        if execution["status"] in ("queued", "created"):
+            if attempt % 3 == 0:  # print on first attempt and thereafter every 3 attempts
+                click.echo(f"Execution #{counter} is {execution['status']}. Waiting for it to start...")
+            time.sleep(5)
+            continue
+        if execution["status"] != "started":
+            raise CLIException(f"Execution #{counter} is {execution['status']}. Cannot SSH into it.")
+        break
+    return call_with_retry(
+        func=lambda: get_ssh_details_from_execution(execution=execution),
+        retries=5,
+        delay_range=(1, 7),
+    )
--- a/tests/commands/execution/test_watch.py
+++ b/tests/commands/execution/test_watch.py
@@ -1,12 +1,8 @@
 import time
 
-from tests.commands.execution.utils import get_execution_data_mock
+from tests.commands.execution.utils import get_execution_data_mock, no_sleep
 from tests.fixture_data import EXECUTION_DATA, PROJECT_DATA
 from valohai_cli.commands.execution.watch import watch
-
-
-def no_sleep(t):
-    raise KeyboardInterrupt("no... sleep... til... Brooklyn!")
 
 
 def test_execution_watch(runner, logged_in_and_linked, monkeypatch):
--- a/tests/commands/execution/utils.py
+++ b/tests/commands/execution/utils.py
@@ -36,3 +36,7 @@ def get_execution_data_mock():
     m.delete(execution_by_counter_url, json={"ok": True})
     m.post(re.compile("^https://app.valohai.com/api/v0/data/(.+?)/purge/$"), json={"ok": True})
     return m
+
+
+def no_sleep(t):
+    raise KeyboardInterrupt("no... sleep... til... Brooklyn!")
--- a/valohai_cli/settings/paths.py
+++ b/valohai_cli/settings/paths.py
@@ -12,6 +12,10 @@ def get_settings_root_path() -> str:  # pragma: no cover
 
 
 def get_settings_file_name(name: str) -> str:
+    """
+    Get the path to a settings file in the user's configuration directory.
+    name can be empty string to get the directory itself.
+    """
     path = os.environ.get("VALOHAI_CONFIG_DIR")
     if path:
         if not os.path.isdir(path):


tarekoraby:
Does this resolve #312?

I assume this is the final piece of [#7172](https://github.com/valohai/roi/issues/7172). If so, please share a screenshot or a recording of the typical steps the end user will experience after merging this PR.

valohai_cli/utils/ssh.py
```diff
@@ -0,0 +1,122 @@
...
+
+    if 1 <= choice <= len(key_pairs):
+        selected_private_key = key_pairs[choice - 1][0]
+        return selected_private_key
+    else:
+        click.echo("Invalid selection Please try again")
+        return None
+
+
+def get_default_ssh_folder() -> str:
```
akx, line 86: We've used the term `directory` before, let's use it here too.

valohai_cli/utils/ssh.py
```diff
@@ -0,0 +1,122 @@
...
+    home_dir = os.path.expanduser("~")
+    return os.path.join(home_dir, ".ssh")
```
akx, line 88:
```suggestion
    return os.path.expanduser("~/.ssh")
```

valohai_cli/utils/ssh.py
```diff
@@ -0,0 +1,122 @@
...
+    choice = click.prompt("Enter the number of the key pair, enter -1 to select none of the above", type=int)
+    if choice == -1:
+        return None
+
+    if 1 <= choice <= len(key_pairs):
+        selected_private_key = key_pairs[choice - 1][0]
+        return selected_private_key
+    else:
+        click.echo("Invalid selection Please try again")
+        return None
```
akx, line 83: Use `valohai_cli.utils.cli_utils.prompt_from_list()` instead :)

valohai_cli/log_manager.py
```diff
@@ -35,3 +36,13 @@ def fetch_events(self, limit: Optional[int] = None) -> dict:
...
+
+    def fetch_status_events(self, first_n: Optional[int] = None) -> dict:
+        params = {}
+        if first_n is not None:
+            params['first_n'] = first_n
+        events_response = request('get', self.status_events_url, params=params).json()
+        return {
+            'total': events_response.get('total'),
+            'status_events': events_response.get('status_events', ()),
+        }
```
akx, line 48: * I'm not sure this should be in `LogManager`, since it doesn't touch `LogManager`'s state (seen messages) at all. It should probably just be a free function.
* Since `total` from the dict isn't used at all, this should maybe just return an `Iterable[dict]`, i.e. `yield from events_response.get('status_events', ())`.

valohai_cli/commands/execution/ssh.py
```diff
@@ -0,0 +1,72 @@
...
+@counter_argument
+@click.option('--private-key-file', default=None, help='Private SSH key to use for the connection.')
+@click.option('--address', default=None, help='Container IP address to connect to.')
+def ssh(counter: str, private_key_file: str, address: str) -> None:
+    """
+    Make SSH Connection to the execution container.
+    """
+    if private_key_file and not os.path.exists(private_key_file):
+        click.echo(f"No such key file found: {private_key_file}")
+        return
```
akx, line 30: This would cause the command to finish with exit code 0 (success).

Raise a `click.UsageError()` instead.

valohai_cli/utils/ssh.py
```diff
@@ -0,0 +1,122 @@
...
+        click.echo(f"Value error: {ve}", err=True)
+    except Exception as ex:
+        click.echo(f"An unexpected error occurred: {ex}", err=True)
+
+
+def find_ssh_details(events: list, prefix: str = "::ssh::") -> Optional[dict]:
+    for event in events:
+        if event['message'].startswith(prefix):
+            return json.loads(event['message'][len(prefix):])
+    return None
```
akx, line 48: * I think this should try to parse the `::ssh::` dict into either a tuple (host, port) or a `dataclass` designed for that purpose.
* The `prefix` doesn't IMO need to be configurable – how would we find SSH details from a non-`::ssh::` message?

valohai_cli/commands/execution/ssh.py
```diff
@@ -0,0 +1,72 @@
...
+        make_ssh_connection_to_container(address, port, private_key_file)
+        return
+    else:
+        for folder in [CLI_CONFIG_DIR, DEFAULT_SSH_DIR]:
+            click.echo(f"Looking for private key in {folder}...")
+            selected_private_key = select_key_from_folder(folder)
+            if selected_private_key:
+                make_ssh_connection_to_container(address, port, selected_private_key)
+                return
+    click.echo(NO_KEY_FOUND_MESSAGE)
```
akx, line 72: This too should be an error condition, not just an echo.

valohai_cli/commands/execution/ssh.py
```diff
@@ -0,0 +1,72 @@
...
+    select_key_from_folder,
+)
+
+DEFAULT_SSH_DIR = get_default_ssh_folder()
+
+
+@click.command()
+@counter_argument
+@click.option('--private-key-file', default=None, help='Private SSH key to use for the connection.')
+@click.option('--address', default=None, help='Container IP address to connect to.')
```
akx, line 23: The help text should specify the format required (looking at the code, it's "address:port", not just an IP address).

valohai_cli/commands/execution/ssh.py
```diff
@@ -0,0 +1,72 @@
...
+        lm = LogManager(execution)
+        events_response = lm.fetch_events(limit=100)
+        events = events_response['events']
+        ssh_details = find_ssh_details(events)
+    else:
+        ip, port = address.split(":")
+        ssh_details = {
+            'address': ip,
+            'port': port
+        }
```
akx, line 51:
```suggestion
        ip_address, _, port = address.partition(":")
        ssh_details = {
            "address": ip_address,
            "port": int(port),
        }
```

akx, line 51: Feels like the entire "get project, get execution, parse SSH details" bit should be moved to a free helper function.

Then we wouldn't even need to retrieve the project and execution at all if the user already provided an address to SSH to.

valohai_cli/commands/execution/ssh.py
```diff
@@ -0,0 +1,72 @@
...
+        click.echo(f"Using provided private key file: {private_key_file}")
+        make_ssh_connection_to_container(address, port, private_key_file)
+        return
+    else:
+        for folder in [CLI_CONFIG_DIR, DEFAULT_SSH_DIR]:
+            click.echo(f"Looking for private key in {folder}...")
+            selected_private_key = select_key_from_folder(folder)
+            if selected_private_key:
+                make_ssh_connection_to_container(address, port, selected_private_key)
+                return
```
akx, line 71: Feels like there should be only one call to `make_ssh_connection_to_container()` in this function 🤔

valohai_cli/utils/ssh.py
```diff
@@ -0,0 +1,122 @@
...
+            'ssh',
+            '-i', private_key_path,
+            f'root@{address}',
+            '-p', port,
+            '-t', '/bin/bash'
+        ]
+        ssh_command = [
+            'ssh',
+            '-t', 'memona008@github.com'
+        ]
```
akx, line 28: 🤔 😄 

```suggestion
```

valohai_cli/commands/execution/ssh.py
akx, line 71: * The `for folder ... find key...` loop should probably also be in `utils/ssh.py`.
* We should probably find all keypairs in all of the given directories, and prompt the user only once. Otherwise, it's a bit silly that you get two "do you want to use this key?" prompts..?

valohai_cli/utils/ssh.py
```diff
@@ -0,0 +1,122 @@
...
+        result = subprocess.run(ssh_command, check=True, text=True, capture_output=True)
+        click.echo(result.stdout)
+
+    except subprocess.CalledProcessError as e:
+        click.echo(f"Failed to execute SSH command: {e}", err=True)
+        if e.stderr:
+            click.echo(f"Error details: {e.stderr}", err=True)
+    except FileNotFoundError:
+        click.echo("SSH command not found. Ensure SSH is installed and available in your PATH.", err=True)
```
akx, line 37: Maybe use `shutil.which()` to find `ssh` and raise if it returns None. (Otherwise, use the return value in the subprocess command.)

valohai_cli/utils/ssh.py
```diff
@@ -0,0 +1,122 @@
...
+    except ValueError as ve:
+        click.echo(f"Value error: {ve}", err=True)
+    except Exception as ex:
+        click.echo(f"An unexpected error occurred: {ex}", err=True)
```
akx, line 41: We shouldn't be catching these, I don't think?

valohai_cli/utils/ssh.py
```diff
@@ -0,0 +1,122 @@
...
+from valohai_cli.settings.paths import get_settings_root_path
+
+CLI_CONFIG_DIR = get_settings_root_path() + "valohai-cli/"
+NO_KEY_FOUND_MESSAGE = "No SSH key pairs found. Please try to generate new SSH key pair to connect."
+
+
+def make_ssh_connection_to_container(address: str, port: str, private_key_path: str) -> None:
+    try:
+        if not all([address, port, private_key_path]):
+            raise ValueError("Missing necessary SSH connection information.")
```
akx, line 17: Doesn't the caller validate all of these?

valohai_cli/utils/ssh.py
```diff
@@ -0,0 +1,122 @@
...
+
+import click
+
+from valohai_cli.settings.paths import get_settings_root_path
+
+CLI_CONFIG_DIR = get_settings_root_path() + "valohai-cli/"
+NO_KEY_FOUND_MESSAGE = "No SSH key pairs found. Please try to generate new SSH key pair to connect."
+
+
+def make_ssh_connection_to_container(address: str, port: str, private_key_path: str) -> None:
```
akx, line 14: `port` should be an integer.

valohai_cli/utils/ssh.py
```diff
@@ -0,0 +1,122 @@
...
+            f'root@{address}',
+            '-p', port,
+            '-t', '/bin/bash'
+        ]
+        ssh_command = [
+            'ssh',
+            '-t', 'memona008@github.com'
+        ]
+        result = subprocess.run(ssh_command, check=True, text=True, capture_output=True)
+        click.echo(result.stdout)
```
akx, line 30: I don't think this is what we want if we want to start an interactive SSH session...

valohai_cli/utils/ssh.py
```diff
@@ -0,0 +1,122 @@
...
+        click.echo("Invalid selection Please try again")
+        return None
+
+
+def get_default_ssh_folder() -> str:
+    home_dir = os.path.expanduser("~")
+    return os.path.join(home_dir, ".ssh")
+
+
+def get_ssh_key_pairs_from_dir(folder_path: str) -> tuple:
```
akx, line 91: * This too should probably return a generator of some `Keypair` dataclass, to avoid `key_pairs[0][0]`.
* `directory` instead of `folder`

valohai_cli/utils/ssh.py
```diff
@@ -0,0 +1,122 @@
...
+    # Match private keys with public keys
+    for base_name, private_key in private_keys.items():
+        base_name = base_name.split(".")[0]
+        if base_name in public_keys:
+            key_pairs.append((private_key, public_keys[base_name]))
+
+    if key_pairs:
+        return True, key_pairs
+    else:
+        return False, []
```
akx, line 122: Since we kind of always need the public part to be around (to eventually send with the execution configuration), I think we could just try to find `.pub`s, then assume they'll be paired with a suffixless private key part. Also, let's not support `ssh-dss` keys, they're deprecated.

Something like:

```suggestion
    for pth in pathlib.Path(folder_path).glob("*.pub"):
        if not pth.is_file():
            continue
        try:
            with pth.open("rb") as f:
                if not first_line.startswith((b"ssh-rsa", b"ecdsa-sha2-", b"ssh-ed25519")):
                    continue
            priv_pth = pth.with_suffix("")
            if priv_pth.is_file():
                yield Keypair(public_path=pth, private_path=priv_pth)
        except OSError:  # e.g. unreadable key
            pass
```

valohai_cli/utils/ssh.py
```diff
@@ -0,0 +1,122 @@
...
+            f"Do you want to use {key_pairs[0][0]} as private key for SSH connection?"):
+            selected_private_key = key_pairs[0][0]
+        elif ssh_pair_found and len(key_pairs) > 1:
+            selected_private_key = prompt_for_key_selection(key_pairs)
+        else:
+            selected_private_key = None
+        return selected_private_key
+    else:
+        click.echo(f"No such directory found: {folder_path}")
+        return None
```
akx, line 64: I think the caller should validate it passes in a proper directory.

valohai_cli/commands/execution/ssh.py
```diff
@@ -0,0 +1,72 @@
...
+from valohai_cli.utils.ssh import (
+    CLI_CONFIG_DIR,
+    NO_KEY_FOUND_MESSAGE,
+    find_ssh_details,
+    get_default_ssh_folder,
+    make_ssh_connection_to_container,
+    select_key_from_folder,
+)
+
+DEFAULT_SSH_DIR = get_default_ssh_folder()
```
akx, line 17: I don't think this needs to be a module-level constant here.

valohai_cli/utils/ssh.py
```diff
@@ -0,0 +1,122 @@
+import json
+import os
+import subprocess
+from typing import Optional
+
+import click
+
+from valohai_cli.settings.paths import get_settings_root_path
+
+CLI_CONFIG_DIR = get_settings_root_path() + "valohai-cli/"
```
akx, line 10: I think this statement should just use `get_settings_file_name()` so it knows about `VALOHAI_CONFIG_DIR`. 

In `get_settings_file_name()`'s docstring, we should maybe clarify that it's OK to e.g. pass in an empty string for `name`...

Finally, I don't think this needs to be a module-level constant either, if it'll only get used eventually in the "find some SSH key" loop.

akx (changes requested):

memona008:
> If so, please share a screenshot or a recording of the typical steps the end user will experience after merging this PR.

Thanks for the reminder. I'll do it once all the comments are addressed and PR is in its final shape so that our CS would get the latest and final recording/screenshot/steps.

valohai_cli/utils/ssh.py
memona008: aaahh noooooo waayy i pushed the testing details *dying with embarassment*

valohai_cli/commands/execution/run/dynamic_run_command.py
```diff
@@ -59,6 +61,7 @@ def __init__(
         tags: Optional[Sequence[str]] = None,
         runtime_config: Optional[dict] = None,
         runtime_config_preset: Optional[str] = None,
+        ssh: Optional[bool] = False,
```
akx, line 64: `Optional[bool]` implies `None` is an option here. Is it?

valohai_cli/commands/execution/run/dynamic_run_command.py
```diff
@@ -196,6 +201,13 @@ def execute(self, **kwargs: Any) -> None:
                 download_directory=self.download_directory,
             )
 
+        if self.ssh:
+            time.sleep(10)
```
akx, line 205: This seems awfully arbitrary. If the idea is to wait until we might be able to see the SSH details, the `ssh` command should instead have an option to loop until it finds the SSH details.

valohai_cli/commands/execution/run/dynamic_run_command.py
```diff
@@ -196,6 +201,13 @@ def execute(self, **kwargs: Any) -> None:
                 download_directory=self.download_directory,
             )
 
+        if self.ssh:
```
akx, line 204: Since the `ssh` command will block the terminal (interactive SSH session and all), this should occur after the `open_browser` `if`, so you can open a browser, then wait for SSH. Otherwise you'd get an SSH session, and after you ctrl+c out of it, a browser would pop up.

valohai_cli/commands/execution/run/frontend_command.py
```diff
@@ -184,6 +189,7 @@ def run(
     k8s_devices: List[str],
     k8s_device_none: bool,
     k8s_preset: Optional[str],
+    ssh: Optional[bool] = False,
```
akx, line 192: `ssh`, `sync` and `watch` are mutually exclusive (all of them are blocking processes), please make it so :)

akx: I doubt `None` is an option here:
```suggestion
    ssh: bool = False,
```

valohai_cli/commands/execution/ssh.py
```diff
@@ -0,0 +1,47 @@
...
+)
+
+
+@click.command()
+@counter_argument
+@click.option("--private-key-file", default=None, help="Private SSH key to use for the connection.")
+@click.option(
+    "--address",
+    default=None,
+    help='Address of the container in "ip:port" format.If not provided, '
```
akx, line 20:
```suggestion
    help='Address of the container in "ip:port" format. If not provided, '
```

valohai_cli/utils/ssh.py
```diff
@@ -0,0 +1,131 @@
...
+
+
+@dataclass
+class Keypair:
+    def __init__(self, public_path: pathlib.Path, private_path: pathlib.Path) -> None:
+        self.public_path = public_path
+        self.private_path = private_path
+
+    def __repr__(self) -> str:
+        return f"Keypair(public_path={self.public_path}, private_path={self.private_path})"
```
akx, line 24: [Dataclasses auto-generate these for you.](https://docs.python.org/3.12/library/dataclasses.html#module-dataclasses)

```suggestion
    public_path: pathlib.Path
    private_path: pathlib.Path
```

valohai_cli/utils/ssh.py
```diff
@@ -0,0 +1,131 @@
...
+
+import click
+
+from valohai_cli.api import request
+from valohai_cli.ctx import get_project
+from valohai_cli.settings import get_settings_file_name
+from valohai_cli.utils.cli_utils import prompt_from_list
+
+
+@dataclass
```
akx, line 17: Might as well make it immutable:
```suggestion
@dataclass(frozen=True)
```

valohai_cli/utils/ssh.py
```diff
@@ -0,0 +1,131 @@
...
+        return f"Keypair(public_path={self.public_path}, private_path={self.private_path})"
+
+
+def make_ssh_connection_to_container(address: str, port: int, private_key_path: str) -> None:
+    try:
+        ssh_path = shutil.which("ssh")
+        if ssh_path is None:
+            raise FileNotFoundError(
+                "SSH command not found. Ensure SSH is installed and available in your PATH.",
+            )
```
akx, line 33: This can be outside the `try:` block.

valohai_cli/utils/ssh.py
```diff
@@ -0,0 +1,131 @@
...
+            str(port),
+            "-t",
+            "/bin/bash",
+        ]
+        subprocess.run(ssh_command)
+
+    except subprocess.CalledProcessError as e:
+        click.echo(f"Failed to execute SSH command: {e}", err=True)
+        if e.stderr:
+            click.echo(f"Error details: {e.stderr}", err=True)
```
akx, line 50: I don't think this is (easily) reachable (and the `stderr` bit especially isn't, since you're not capturing `stderr`). I'm not sure there's much value in the whole `try:except:` here; the user would see the error from `ssh`, wouldn't they?

valohai_cli/utils/ssh.py
```diff
@@ -0,0 +1,131 @@
...
+        except OSError:  # e.g. unreadable key
+            pass
+
+
+def select_private_key_from_possible_directories() -> Optional[str]:
+    default_ssh_dir = get_default_ssh_directory()
+    cli_config_dir = get_settings_file_name("")
+    all_keypairs: list[Keypair] = []
+    for dir in [cli_config_dir, default_ssh_dir]:
+        click.echo(f"Looking for private key in {dir}...")
```
akx, line 89: Is this verbosity necessary? Maybe we should print the directories only if we didn't find any keys.

valohai_cli/utils/ssh.py
```diff
@@ -0,0 +1,131 @@
...
+    project = get_project(require=True)
+    assert project
+    execution = project.get_execution_from_counter(
+        counter=counter,
+        params={
+            "exclude": "metadata, events",
+        },
+    )
+    events_response = fetch_status_events(execution, first_n=100)
+    ssh_details = find_ssh_details(list(events_response))
```
akx, line 128:
```suggestion
    ssh_details = find_ssh_details(fetch_status_events(execution, first_n=100))
```
(and make the argument type accepted by `find_ssh_details` an `Iterable`)

valohai_cli/commands/execution/run/dynamic_run_command.py
```diff
@@ -75,6 +78,7 @@ def __init__(
         :param download_directory: Where to (if somewhere) to download execution outputs (sync mode)
         :param runtime_config: Runtime config dict
         :param runtime_config_preset: Runtime config preset identifier (UUID)
+        :param ssh: Whether opens an interactive ssh connection for debugging
```
akx, line 81:
```suggestion
        :param ssh: Whether to chain to `exec ssh` afterward
```
(to copy the documentation for `watch`)

akx (changes requested):

valohai_cli/commands/execution/ssh.py
```diff
@@ -0,0 +1,49 @@
...
+                raise CLIException(f"Invalid address format: {e}")
+        else:
+            ip_address, port = get_ssh_details_with_retry(counter)
+
+        click.echo(f"SSH Container address is {ip_address}:{port}")
+        if selected_private_key := private_key_file or select_private_key_from_possible_directories():
+            make_ssh_connection_to_container(ip_address, port, selected_private_key)
+    except Exception as e:
+        click.secho(str(e), fg="red", err=True)
+        raise click.Abort()
```
akx, line 49: What's the benefit of having everything wrapped in a try/except here? 🤔 Since `CLIException`s derive from `click.ClickException`, they'll be handled nicely anyway (see https://click.palletsprojects.com/en/8.1.x/exceptions/).

valohai_cli/utils/ssh.py
```diff
@@ -0,0 +1,144 @@
...
+    if not ssh_details:
+        raise click.UsageError("No SSH details found")
+    return ssh_details
+
+
+def get_ssh_details_with_retry(
+    counter: int,
+    retries: int = 5,
+    delay_range: Tuple[int, int] = (1, 7),
```
akx, line 129: Since you never pass these in to `get_ssh_details_with_retry`, they maybe don't need to be arguments to this function.

valohai_cli/utils/__init__.py
```diff
@@ -154,3 +155,25 @@ def parse_environment_variable_strings(
...
+) -> T:
+    kwargs = kwargs or {}
+    for attempt in range(retries):
+        try:
+            return func(*args, **kwargs)
+        except exception as e:
+            click.echo(f"{attempt + 1}/{retries} Retrying: {e}...")
+            time.sleep(random.randint(*delay_range))
+
+    raise exception(f"Failed after {retries} retries")
```
akx, line 179: * We don't need `args` and `kwargs` here; `func` itself could be a `lambda` or `functools.partial()`.
* We could want to catch multiple exception classes (e.g. `(IOError, TimeoutError)`)
* `raise exception` is a bit weird – you can't necessarily be certain that any given exception type passed in by the user is constructible with a string. It would also be useful to be able to access the exception.

Maybe something like:

```suggestion
T = TypeVar("T")


def call_with_retry(
    func: Callable[[], T],
    retries: int = 3,
    delay_range: Tuple[int, int] = (1, 5),
    retry_on_exception_classes: Tuple[Type[Exception], ...] = (Exception,),
) -> T:
    for attempt in range(retries):
        try:
            return func()
        except retry_on_exception_classes as e:
            if attempt == retries - 1:
                raise RuntimeError(f"Failed after {retries} attempts") from e
            click.echo(f"{attempt + 1}/{retries} Retrying: {e}...")
            time.sleep(random.uniform(*delay_range))
```

valohai_cli/commands/execution/run/frontend_command.py
```diff
@@ -184,6 +189,7 @@ def run(
     k8s_devices: List[str],
     k8s_device_none: bool,
     k8s_preset: Optional[str],
+    ssh: bool = False,
```
akx, line 192: As said in the last review:

> ssh, sync and watch are mutually exclusive (all of them are blocking processes), please make it so :)

There is also precedent for 2 mutually exclusive options (see `if download_directory and watch:`).

valohai_cli/commands/execution/ssh.py
```diff
@@ -0,0 +1,49 @@
...
+from valohai_cli.utils.ssh import (
+    get_ssh_details_with_retry,
+    make_ssh_connection_to_container,
+    select_private_key_from_possible_directories,
+)
+
+
+@click.command()
+@counter_argument
+@click.option("--private-key-file", default=None, help="Private SSH key to use for the connection.")
```
akx, line 16:
```suggestion
@click.option("--private-key-file", default=None, type=click.Path(file_okay=True, exists=True), help="Private SSH key to use for the connection.")
```
and you don't need to validate the existence of the file by hand.

valohai_cli/commands/execution/ssh.py
```diff
@@ -0,0 +1,49 @@
...
+    help='Address of the container in "ip:port" format. If not provided, '
+    "the address from the execution will be used.",
+)
+def ssh(counter: int, private_key_file: str, address: str) -> None:
+    """
+    Make SSH Connection to the execution container.
+    """
+    try:
+        if private_key_file and not os.path.exists(private_key_file):
+            raise click.UsageError(f"No such key file found: {private_key_file}")
```
akx, line 29:
```suggestion
```
✂️ if you use `click.Path()`.

valohai_cli/commands/execution/ssh.py
```diff
@@ -0,0 +1,49 @@
...
+                    raise CLIException("Address must be in 'ip:port' format.")
+                port = int(port_str)
+                if port <= 1023:
+                    raise CLIException("Port must be above 1023")
+            except ValueError as e:
+                raise CLIException(f"Invalid address format: {e}")
+        else:
+            ip_address, port = get_ssh_details_with_retry(counter)
+
+        click.echo(f"SSH Container address is {ip_address}:{port}")
```
akx, line 44: "SSH Container address"? 🤔 Sounds like there is a separate container?

valohai_cli/utils/ssh.py
```diff
@@ -0,0 +1,144 @@
...
+from valohai_cli.utils.cli_utils import prompt_from_list
+
+
+@dataclass(frozen=True)
+class Keypair:
+    public_path: pathlib.Path
+    private_path: pathlib.Path
+
+
+def make_ssh_connection_to_container(address: str, port: int, private_key_path: str) -> None:
```
akx, line 24: I don't think there's much value in saying `_to_container` here?
```suggestion
def make_ssh_connection(address: str, port: int, private_key_path: str) -> None:
```

valohai_cli/utils/ssh.py
```diff
@@ -0,0 +1,144 @@
...
+    prefix = "::ssh::"
+    for event in events:
+        if event["message"].startswith(prefix):
+            data = json.loads(event["message"][len(prefix) :])
+            return data["address"], int(data["port"])
+    return None
+
+
+def get_default_ssh_directory() -> str:
+    return os.path.expanduser("~/.ssh")
```
akx, line 53: This probably could be inlined into the single function that uses it now.

valohai_cli/utils/ssh.py
```diff
@@ -0,0 +1,144 @@
...
+            pass
+
+
+def select_private_key_from_possible_directories() -> Optional[str]:
+    default_ssh_dir = get_default_ssh_directory()
+    cli_config_dir = get_settings_file_name("")
+    all_keypairs: list[Keypair] = []
+    all_dirs = [cli_config_dir, default_ssh_dir]
+    for dir in all_dirs:
+        if not os.path.exists(dir):
```
akx, line 80:
```suggestion
        if not os.path.isdir(dir):
```

valohai_cli/utils/ssh.py
```diff
@@ -0,0 +1,144 @@
...
+            continue
+        if key_pairs := get_ssh_key_pairs_from_dir(dir):
+            all_keypairs.extend(key_pairs)
```
akx, line 83: It's a generator, you don't need to check before extending.

```suggestion
        all_keypairs.extend(get_ssh_key_pairs_from_dir(dir))
```

valohai_cli/utils/ssh.py
```diff
@@ -0,0 +1,144 @@
...
+    Fetch SSH details for a given execution, retrying up to `retries` times with a random delay between retries.
+
+    :param counter: The execution counter.
+    :param retries: Number of retries before giving up.
+    :param delay_range: Range for random delay in seconds between retries.
+    :return: A tuple of (IP address, port) for the SSH connection.
+    """
+    return call_with_retry(
+        func=get_ssh_details_from_execution,
+        args=(counter,),
```
akx, line 141: As discussed around simplifying `call_with_retry`,
```suggestion
        func=partial(get_ssh_details_from_execution, counter=counter),
```
or
```suggestion
        func=lambda: get_ssh_details_from_execution(counter),
```

akx (changes requested):

valohai_cli/models/project.py
```diff
@@ -70,6 +70,7 @@ def get_execution_from_counter(
         self,
         counter: Union[int, str],
         params: Optional[dict] = None,
+        raise_on_not_running: bool = False,
```
akx, line 73: I don't think this functionality belongs in `get_execution_from_counter()`.

You could have a ssh.py-specific function that `get_execution_from_counter()`s and raises on a given status type, but also see the commentary in that module.

valohai_cli/models/project.py
```diff
@@ -84,6 +85,11 @@ def get_execution_from_counter(
                 params=(params or {}),
             ).json()
             assert isinstance(data, dict)
+
+            if raise_on_not_running and data["status"] == "queued":
```
akx, line 89: So the new parameter doesn't do what it's named to do? Based on this it'd be `raise_if_queued`?

Either way, for the "wait until it's up and running" functionality for SSH you really would also want to raise if the execution has e.g. stopped or errored already.

valohai_cli/models/project.py
```diff
@@ -84,6 +85,11 @@ def get_execution_from_counter(
...
+                raise ExecutionNotRunning(
+                    f"Execution #{counter} is not running. It's current status is: {data['status']}",
```
akx, line 91: Typo: `it's` -> `its`.
```suggestion
                    f"Execution #{counter} is not running. Its current status is: {data['status']}",
```

Also, you're ensuring `status` is queued above, so there's no way this could ever be "Its current status is: queued"...

valohai_cli/utils/ssh.py
```diff
@@ -0,0 +1,147 @@
...
+    if not ssh_details:
+        raise click.UsageError("No SSH details found")
+    return ssh_details
+
+
+def get_ssh_details_with_retry(
+    counter: int,
+) -> Tuple[str, int]:
+    """
+    Fetch SSH details for a given execution, retrying up to `retries` times with a random delay between retries.
```
akx, line 137: The docstring is referring to now-deleted arguments...
```suggestion
    Fetch SSH details for a given execution.
```

valohai_cli/utils/ssh.py
```diff
@@ -0,0 +1,147 @@
...
+            counter=counter,
+            params={
+                "exclude": "metadata, events",
+            },
+        ),
+        retry_on_exception_classes=(ExecutionNotRunning,),
+        retries=5,
+        delay_range=(1, 7),
+    )
+    return execution
```
akx, line 122: I don't think this needs to be a "public" function if it only is ever called by `get_ssh_details_with_retry`.

Also, I'm not sure `call_with_retry()` is a great idea for waiting for the execution to become `started` – you'd be waiting for 5 to 35 seconds, and then giving up altogether.

I think it'd be better to:

* loop forever until the execution becomes `started` (and error out if it's e.g. stopped or errored; we can't SSH there)
  * maybe 5 seconds between retries?
  * print out a "Waiting for execution to start" after the first few tries and thereafter every now and then
  * the user can always ctrl+c out of this
* then in a subsequent loop or retry-loop, find the SSH events.

valohai_cli/commands/execution/ssh.py
```diff
@@ -0,0 +1,45 @@
...
+            if port <= 1023:
+                raise CLIException("Port must be above 1023")
+        except ValueError as e:
+            raise CLIException(f"Invalid address format: {e}")
+    else:
+        ip_address, port = get_ssh_details_with_retry(counter)
+
+    click.echo(f"SSH address is {ip_address}:{port}")
+    if selected_private_key := private_key_file or select_private_key_from_possible_directories():
+        make_ssh_connection(ip_address, port, selected_private_key)
```
akx, line 45: I think this reads more clearly as
```suggestion
    if not private_key_file:
        private_key_file = select_private_key_from_possible_directories()
    make_ssh_connection(ip_address, port, private_key_file)
```

valohai_cli/utils/ssh.py
```diff
@@ -0,0 +1,147 @@
...
+                    private_key_paths = [pth.with_suffix(""), pth.with_suffix(".pem")]
+                    for private_key_path in private_key_paths:
+                        if private_key_path.is_file():
+                            yield Keypair(public_path=pth, private_path=private_key_path)
+                            break
+        except OSError:  # e.g. unreadable key
+            pass
+
+
+def select_private_key_from_possible_directories() -> Optional[str]:
```
akx, line 71: Isn't the rtype always `str`? This either raises or returns a string.
```suggestion
def select_private_key_from_possible_directories() -> str:
```

akx (changes requested):

valohai_cli/utils/ssh.py
```diff
@@ -0,0 +1,159 @@
...
+
+    execution = project.get_execution_from_counter(
+        counter=counter,
+        params={
+            "exclude": "metadata, events",
+        },
+    )
+    if execution["status"] in ("queued", "created"):
+        raise ExecutionQueued(f"Execution #{counter} is queued. Waiting for it to start...")
+    if execution["status"] in ("stopped", "stopping (hard)", "stopping", "error", "complete"):
```
akx, line 119: The `status` from the API will not be `"stopping (hard)"`, it will be `"stopping_hard"`...

valohai_cli/exceptions.py
```diff
@@ -127,3 +127,11 @@ class NoCommit(ValueError, CLIException):
...
         self.directory = directory
         super().__init__(f"{directory} has no commits")
+
+
+class ExecutionQueued(CLIException):
+    color = "yellow"
+
+
+class ExecutionFinished(CLIException):
+    color = "yellow"
```
akx, line 137: Since these are only used by the `ssh` machinery, and only internally therein, I don't think they should be here. (But see more comments in the `ssh.py` module, I don't think these are really necessary at all...)

valohai_cli/utils/ssh.py
```diff
@@ -0,0 +1,159 @@
...
+            execution = _get_execution_from_counter(counter)
+            break
+        except Exception as e:
+            if isinstance(e, ExecutionQueued):
+                echo_attempts -= 1
+                if echo_attempts > 0:
+                    click.echo(str(e))
+                time.sleep(5)
+                continue
+            raise
```
akx, line 153: I would inline `_get_execution_from_counter()` here and avoid the extra exception classes. `count()` is from `itertools`.

```suggestion
    project = get_project(require=True)
    assert project

    for attempt in count():
        execution = project.get_execution_from_counter(
            counter=counter,
            params={"exclude": "metadata, events"},
        )
        if execution["status"] in ("queued", "created"):
            if attempt % 3 == 0:  # print on first attempt and thereafter every 3 attempts
                click.echo(f"Execution #{counter} is {execution['status']}. Waiting for it to start...")
            time.sleep(5)
            continue
        if execution["status"] != "started":
            raise CLIException(f"Execution #{counter} is {execution['status']}. Cannot SSH into it.")
```

akx (changes requested):
Still a bit to go, sorry...

akx (changes requested):
Looks good, but could we have some tests for this?

tests/commands/execution/test_ssh.py
```diff
@@ -0,0 +1,68 @@
...
+    EXECUTION_DATA["status"] = "queued"
+    monkeypatch.setattr(time, "sleep", no_sleep)
+    with get_execution_data_mock():
+        output = runner.invoke(ssh, [str(counter)], catch_exceptions=False).output
+        assert f"Execution #{counter} is queued. Waiting for it to start...\n" in output
+
+
+def test_ssh_with_no_ssh_details_present(runner, logged_in_and_linked, monkeypatch):
+    counter = EXECUTION_DATA["counter"]
+    EXECUTION_DATA["status"] = "started"
```
akx, line 32: Ditto about modifying `EXECUTION_DATA`.

tests/commands/execution/test_ssh.py
```diff
@@ -0,0 +1,68 @@
...
+        assert "1/5 Retrying: No SSH details found...\n" in output
+        assert "2/5 Retrying: No SSH details found...\n" in output
+        assert "3/5 Retrying: No SSH details found...\n" in output
+        assert "4/5 Retrying: No SSH details found...\n" in output
+        assert "5/5 Retrying: No SSH details found...\n" in output
+
+
+def test_ssh(runner, logged_in_and_linked, monkeypatch):
+    counter = EXECUTION_DATA["counter"]
+    EXECUTION_DATA["status"] = "started"
```
akx, line 46: Ditto about modifying `EXECUTION_DATA`.

tests/commands/execution/test_ssh.py
```diff
@@ -0,0 +1,68 @@
+import pathlib
+import subprocess
+import time
+
+from tests.commands.execution.test_watch import no_sleep
```
akx, line 5: Shouldn't this import from `utils` (since you move it there)?

tests/commands/execution/test_ssh.py
```diff
@@ -0,0 +1,68 @@
...
+        assert f"Error: Execution #{counter} is complete. Cannot SSH into it.\n" in output
+
+        EXECUTION_DATA["status"] = "queued"
+        monkeypatch.setattr(time, "sleep", no_sleep)
+
+
+def test_ssh_in_queued_execution(runner, logged_in_and_linked, monkeypatch):
+    counter = EXECUTION_DATA["counter"]
+    EXECUTION_DATA["status"] = "queued"
+    monkeypatch.setattr(time, "sleep", no_sleep)
```
akx, line 24: Hmm... `no_sleep` always raises a KeyboardInterrupt (i.e. as if the user had interrupted the sleep immediately).

What is it that we're testing here?

tests/commands/execution/test_ssh.py
```diff
@@ -0,0 +1,68 @@
...
+        assert "2/5 Retrying: No SSH details found...\n" in output
+        assert "3/5 Retrying: No SSH details found...\n" in output
+        assert "4/5 Retrying: No SSH details found...\n" in output
+        assert "5/5 Retrying: No SSH details found...\n" in output
+
+
+def test_ssh(runner, logged_in_and_linked, monkeypatch):
+    counter = EXECUTION_DATA["counter"]
+    EXECUTION_DATA["status"] = "started"
+    monkeypatch.setattr(time, "sleep", lambda x: None)
```
akx, line 47: Why does `sleep` need to be mocked here? 🤔

tests/commands/execution/test_ssh.py
```diff
@@ -0,0 +1,68 @@
...
+
+    def mock_prompt():
+        return pathlib.Path("/path/to/private")
```
akx, line 50: Add the `tmp_path` fixture to the test and return that instead of hard-coding a path:
```suggestion
    def mock_prompt():
        return tmp_path
```

tests/fixture_data.py
```diff
@@ -665,3 +664,19 @@ def main(config) -> Pipeline:
...
+            "message": '::ssh::{"port": 2222, "address": "127.0.0.1"}',
+            "time": "2024-09-04T12:16:20.722000",
+        },
+        {
+            "stream": "status",
+            "message": "   $ ssh -i <path-to-private-key> 127.0.0.1 -p 2222 -t /bin/bash",
+            "time": "2024-09-04T12:16:20.723000",
+        },
+    ],
+}
```
akx, line 682: This seems specific to the SSH tests – maybe do like `test_execution_outputs` does and modify the mock `with get_execution_data_mock() as m:` to only return these when they should be present? That would also make `test_ssh_with_no_ssh_details_present` test more actual behavior, as you wouldn't need to mock `find_ssh_details` to a no-op.

tests/commands/execution/test_ssh.py
```diff
@@ -0,0 +1,68 @@
...
+
+    with get_execution_data_mock():
+        output = runner.invoke(ssh, [str(counter)], catch_exceptions=False).output
+        assert "SSH address is 127.0.0.1:2222" in output
+
+        def mock_ssh_connection(*args, **kwargs):
+            print("Mock SSH interactive session executed")
+            return subprocess.CompletedProcess(args=args, returncode=0, stderr=b"")
+
+        monkeypatch.setattr("valohai_cli.commands.execution.ssh.make_ssh_connection", mock_ssh_connection)
```
akx, line 65: Maybe monkeypatch `subprocess.run` instead with a mock, and check that the mock gets called with the correct command line?

tests/commands/execution/test_ssh.py
```diff
@@ -0,0 +1,68 @@
...
+        return pathlib.Path("/path/to/private")
+
+    monkeypatch.setattr(
+        "valohai_cli.commands.execution.ssh.select_private_key_from_possible_directories",
+        mock_prompt,
+    )
+
+    with get_execution_data_mock():
+        output = runner.invoke(ssh, [str(counter)], catch_exceptions=False).output
+        assert "SSH address is 127.0.0.1:2222" in output
```
akx, line 59: I'm not following why this test does two invocations..?

tests/commands/execution/test_ssh.py
```diff
@@ -0,0 +1,68 @@
...
+from tests.commands.execution.utils import get_execution_data_mock
+from tests.fixture_data import EXECUTION_DATA
+from valohai_cli.commands.execution.ssh import ssh
+
+
+def test_ssh_in_completed_execution(runner, logged_in_and_linked, monkeypatch):
+    with get_execution_data_mock():
+        counter = EXECUTION_DATA["counter"]
+        output = runner.invoke(ssh, [str(counter)], catch_exceptions=False).output
+        assert f"Error: Execution #{counter} is complete. Cannot SSH into it.\n" in output
```
akx, line 15: I wouldn't personally test against the full error message, newline and all, but it's okay.

tests/commands/execution/test_ssh.py
```diff
@@ -0,0 +1,68 @@
...
+
```
akx, line 16: Please assert that `output.exit_code` is the expected non-zero status (consider this repeated for all of these error tests).

akx (changes requested):

tests/commands/execution/test_ssh.py
akx: Why would it wait for 5 seconds somewhere to begin with? I thought we only sleep when retrying?

memona008: Oh yes, i added it in the failing scenerios to avoid waiting 
it doesnt make sense here in all-good-flow so removing it 👍🏼

akx (approved):
Thanks for putting up with all of the review comments, I think we ended up with a pretty elegant solution :)

PR #318, opened 2024-11-04
Dependency upgrades & minor CI renovation

akx:
* Ensure that users get new versions of transitive deps
* Upgrade lint tools & fix things so they don't complain
* Speed up CI by using `uv`
* Fix Codecov reporting (token was missing; that had worked before, not so anymore)

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -21,6 +21,13 @@ dependencies = [
     "valohai-yaml>=0.43.0",
 ]
 
+[project.optional-dependencies]
+test = [
+    "pytest-cov~=4.0",
+    "pytest~=7.0",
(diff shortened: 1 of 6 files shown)


pyproject.toml
```diff
@@ -59,6 +59,9 @@ target-version = "py38"
 line-length = 110
 preview = true
 
+[tool.ruff.format]
+quote-style = "preserve"
```
teroyks, line 70: Is this necessary?
Going with uniform quote style everywhere would be preferable IMO.

teroyks: Just for backward compatibility (so every little change doesn’t require reformatting)?

teroyks (approved):
👍

PR #323, opened 2025-02-10
introduce environment_variable_groups to the vh exec run command

Kenneth-Schroeder:
I want to make the environment_variable_groups setting accessible from the CLI and from the valohai yaml.
To achieve this, I have created a PR in valohai-yaml as well ([valohai-yaml#160](https://github.com/valohai/valohai-yaml/pull/160)).

I could test my changes on valohai-cli and the local linting on valohai-yaml, however, the API endpoint seems to validate with an older valohai-yaml version as well, causing an API validation error for the yaml version.


akx (commented):
Mmm, there's a difference between explicitly passing `"environment_variable_groups": []` and not passing it at all to the API.

If you don't pass it at all, the defaults from the project are used; if you explicitly pass an empty list, the defaults are not used.

Eyeballing this, one of those scenarios isn't supported?

Kenneth-Schroeder:
changed default to None, handling None, so all scenarios should be supported :)

valohai_cli/commands/execution/run/dynamic_run_command.py
```diff
@@ -92,6 +93,8 @@ def __init__(
         self.download_directory = download_directory
         self.title = title
         self.environment_variables = dict(environment_variables or {})
+        if environment_variable_groups:
+            self.environment_variable_groups = list(environment_variable_groups)
```
akx, line 97: AISI, this would leave the instance field possibly unset?

```suggestion
        if environment_variable_groups:
            self.environment_variable_groups = list(environment_variable_groups)
        else:
            self.environment_variable_groups = None
```

It doesn't actually matter in this case since we end up doing `getattr(self, "environment_variable_groups", None)`, but it's good hygiene...

akx (approved):

PR #324, opened 2025-02-27
🩹 Pass time-limit YAML setting to execution

teroyks:
Resolves [valohai/roi#8352
](https://github.com/valohai/roi/issues/8352) (`time_limit` not passed to execution via CLI)

The `time-limit` setting is parsed and included in the execution parameters if it is set
(0 = no time limit, so it is not included in the params).

### DIFF
--- a/tests/commands/execution/test_run.py
+++ b/tests/commands/execution/test_run.py
@@ -80,6 +80,31 @@ def test_run_spot_restart(run_test_setup):
     }
 
 
+@pytest.mark.parametrize("time_limit", (None, 121, 0, "10min"))
+def test_run_time_limit(run_test_setup, time_limit):
+    """Make sure that the time-limit setting is correctly passed from YAML config to the API."""
+    step_config = """
+- step:
+    name: Train model
+    image: busybox
+    command: "false"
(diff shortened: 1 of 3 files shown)


teroyks:
Added only as a `valohai.yaml` config parameter for now.
I can add also a command-line option, if needed.

tests/commands/execution/test_run.py
```diff
@@ -80,6 +80,31 @@ def test_run_spot_restart(run_test_setup):
...
+"""
+    if time_limit is not None:
+        step_config += f"    time-limit: {time_limit}\n"
+
+    with open(get_project().get_config_filename(), "w") as yaml_fp:
+        yaml_fp.write(step_config)
+
+    run_test_setup.run()
+    if time_limit:
+        time_limit_in_seconds = 600 if time_limit == "10min" else time_limit
```
akx, line 100: I'd maybe make this a pairwise-passed parametrize parameter (`("time_limit", "expected_time_limit_seconds")` etc.), but this is fine!

akx (approved):
No time like the present!

PR #325, opened 2025-02-28
🔖 Become 0.29.0

teroyks:
- execution priority support
- YAML config time limit support


akx (approved):
Does what it says on the tin!

PR #326, opened 2025-03-05
Allow overriding node environments when starting a pipeline

akx:
Allow setting `--environment` when creating pipelines; it adds/replaces the `environment` value in each node's `template`.


teroyks (approved):
Tested locally with a simple pipeline,
and everything is working as expected!

PR #327, opened 2025-03-05
Become 0.30.0

teroyks:
Support overriding environments for pipeline nodes with `vh pipeline run --environment`.

**Note:** This overrides the environment for **all** the execution and task nodes in the pipeline.

Usable for example when sometimes you need to run a pipeline on more powerful machines than usual.


teroyks:
[Tagged and bagged!](https://github.com/valohai/valohai-cli/releases/tag/v0.30.0)

PR #332, opened 2025-04-15
Handle 'Content-Disposition: attachment' when downloading outputs

akx:
This is required for certain private server configurations.


tokkoro (approved):
Good changes! I wish I used something similar to detect the config on the web UI side. It's looking into the global settings to decide how to handle the download.

Issue #333, opened 2025-05-19
Start a notebook execution from the CLI


teroyks:
It should be possible to start a notebook execution via the CLI (without going to the GUI).

## Command

`vh notebook run`

## API call

[NotebookExecutionCreationCreate](https://staging.valohai.com/api/docs/#operation/NotebookExecutionCreationCreate)

parameters:

- commit
- \* **environment**
- environment_variables
- environment_variable_groups
- runtime_config
- runtime_config_preset
- \* **image**
- inherit_environment_variables
- inputs
- \* **project**
- tags
- title
- create_token
- copy_source
- notebook_datum_ids

teroyks:
Start with the required parameters.
Add more if needed.

PR #334, opened 2025-05-21
Add minimal command to start a notebook execution

teroyks:
`vh notebook run --environment=SLUG_OR_ID --image=PYTHON_IMG`: start a notebook execution.

This is a minimal working version to be able to start the execution: it requires 'environment' and 'image' to be defined as command-line arguments, and otherwise uses default values; there is no customization of the API call yet.

Refs #333 – adding options to customize the run is to be implemented later (after discussing with CS what are the most important ones).


teroyks:
Pushed without tests to check the error were not because of changes to test utils –
test still failing, with no changes to any tested areas.

tests/fixture_data.py
```diff
@@ -258,6 +258,165 @@
...
+        },
+    },
+    "runtime_config": {
+        "autorestart": False,
+        "no_output_timeout": 604800,
+    },
+    "executor_identity": "",
+    "image": "python:3.13",
+    "inputs": [],
+    "interpolated_command": "env PIP_DISABLE_PIP_VERSION_CHECK=1 PIP_ROOT_USER_ACTION=ignore pip install --quiet --no-cache-dir https://dist.valohai.com/valohai-notebook-boot/release/valohai_notebook_boot-2024.1.22.3-py3-none-any.whl && vhnb-boot --server-addr nb.valohai.dev",
```
akx, line 413: These don't need to be specified in such great detail...
```suggestion
    "command": "vhnb-boot --server-addr nb.valohai.dev",
    "environment_variables": {},
    "runtime_config": {
        "autorestart": False,
        "no_output_timeout": 604800,
    },
    "executor_identity": "",
    "image": "python:3.13",
    "inputs": [],
    "interpolated_command": "vhnb-boot --server-addr nb.valohai.dev",
```

akx (changes requested):
Good initial version, tiny change request within.

tests/fixture_data.py
teroyks: Agreed – the response what faithful to the real one returned by the API but most of the stuff is not needed.

akx (approved):
📓 🆗

PR #335, opened 2025-05-22
📝 Format list of supported Python versions as a list

teroyks:
Make the list of supported Python versions more readable by formatting it as an actual list.

This PR was originally just a confirmation that the test issue (fixed in #336) was caused by the CI environment and not code changes. But why not improve the docs while we’re at it!


teroyks:
This was really added mainly to check if the CI test processes are working.

akx (approved):
Listaculous!

PR #336, opened 2025-05-23
Pin Click to latest version that does not break tests

teroyks:
Click v8.2 introduces some changes that break tests, especially where git operations are monkeypatched: it produces the following extra warning output:
```
Warning: The directory is not a Git repository. 
Would you like to just run using the latest commit known by Valohai?
Use latest commit? [Y/n]:Aborted!
```

Naturally, this breaks the tests that check the output for expected test.

Temporarily fixed by pinning `click` to a version below `8.2`.

---

Also update `setup-uv` to `v6` while at it (unrelated, but left there since it was already there due to earlier testing).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -26,7 +26,7 @@ jobs:
       - uses: actions/setup-python@v5
         with:
           python-version: "${{ matrix.python-version }}"
-      - uses: astral-sh/setup-uv@v5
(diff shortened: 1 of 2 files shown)


akx (approved):
Put a pin in it for now!

akx:
(See https://github.com/pallets/click/issues/2522 in Click.)

PR #337, opened 2025-05-23
Add python 3.13 to the test suite

teroyks:
Make sure everything works with the latest official release.
Should not cause anything to break.


teroyks:
While I was updating build stuff anyway…

akx (approved):
test sweet

PR #338, opened 2025-05-23
Project environment variable management commands

akx:
Rehashes @drazendee's PR from 2022, now with rebasing and tests!

Closes #251 (supersedes it)

PR #339, opened 2025-05-23
Become 0.32.0

teroyks:
## Features

- Add support for starting a notebook execution

## Internal Updates

- Add Python 3.13 to the test matrix
- Upgrade CI to `setup-uv@v6`
- Pin `click` to a version that doesn't break the tests


teroyks:
> Maybe let's get #338 in here too? 

Sure, we can handle that first thing Monday morning to get it in!

PR #340, opened 2025-05-26
notebook run: add support for environment variables

akx:
Follow-up on #334 to make it one smidgeon less minimal 😄


teroyks (approved):
It work! 👍

PR #343, opened 2025-06-03
✨ Make the `--environment` parameter optional

### LINKED ISSUE #342: Make the `--environment` argument optional
If an explicit environment argument is not given, instruct the API to use the project’s default environment instead.

teroyks:
If no environment is explicitly specified, instruct the API to use the project default environment.
Works with Roi `20250604.103429Z-93db42506` or newer.

Resolves #342


ruksi (approved):
it does and tests what it says on the tin

lgtm!

PR #346, opened 2025-07-03
✨ Add `--version` option to the `vh` command

teroyks:
The purpose of this option is to make it easier to check which version of valohai-cli you are using (e.g. when debugging stuff).

Usage:

```shell
$ vh --version
vh 0.33.0
```


akx (approved):
Clever!

PR #348, opened 2025-07-30
Get output files with pagination

### LINKED ISSUE #347: Increase the limit for output downloads
Currently there is [a limit of 9000 files for downloading outputs](https://github.com/valohai/valohai-cli/blob/87baae014b5a2122e758631a70fc9964fd4344b9/valohai_cli/commands/execution/outputs.py#L36). In some cases the users have more than 15k outputs they want to download and then end up missing a large chunk of the files. 

It would be great to be able to configure the limit. In this case it would also be good to have some warning for the users that increasing the limit to very large values can cause issues (assumably for example timeouts). 

Moreover, it might be good to increase the limit to something like 20k for now.

akx:
Fixes #347


teroyks (approved):
Should now work for downloads of any size!

Should probably be rebased and updated after merging #349,
to update the code to same Python 3.9(+) format as the rest.

PR #349, opened 2025-07-30
Drop Python 3.8 support

akx:
Python 3.8 has been EOL for 9 months. 

Previous versions of valohai-cli will of course continue to work with Python 3.8.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -23,15 +23,12 @@ jobs:
           - "3.13"
     steps:
       - uses: actions/checkout@v4
-      - uses: actions/setup-python@v5
-        with:
-          python-version: "${{ matrix.python-version }}"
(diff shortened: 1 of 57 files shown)


teroyks (approved):
Slowly moving towards modern code! 👍 
(Even borrowing more modernity from the future!)

teroyks:
@akx for some reason, there are still a couple of Ubuntu 22 checks pending, even though that was removed from the test matrix. Leftovers from some earlier push?

akx:
> @akx for some reason, there are still a couple of Ubuntu 22 checks pending, even though that was removed from the test matrix. Leftovers from some earlier push?

Ah, no, it's just that the "required checks" specified in the repo settings refer to versions too (which makes sense), so those are left pending until I :cattunk: the settings

PR #358, opened 2025-11-25
CLI docs fixes

### LINKED ISSUE #357: Fix missing/ambiguous documentation
As noted by Claude for the new docs:

- N/A description for --debug-port and --debug-key-file - these could use better descriptions
- --all option has None as default instead of False for a boolean
- typos in "resouce" should be "resource" (appears twice)
- Clarify the "STEP-OPTIONS" and "PIPELINE-OPTIONS" - These appear in usage strings (lines 210, 405) but it's not immediately clear what format they take. A brief explanation would help.
- The --title option incorrectly shows -c as its short option

akx:
Fixes #357.

### DIFF
--- a/valohai_cli/commands/execution/run/frontend_command.py
+++ b/valohai_cli/commands/execution/run/frontend_command.py
@@ -40,7 +40,7 @@ EMPTY_DICT_PLACEHOLDER = object()
     "--environment",
     "-e",
     default=None,
-    help='The environment UUID or slug to use (see "vh env")',
+    help='The environment UUID or slug to use (see "vh environment list")',
 )
 @click.option(
     "--environment-variable-groups",


noorula (approved):
Lgtm! Are these automatically added to the docs by the way or does that require another update?

valohai_cli/commands/pipeline/run/run.py
```diff
@@ -33,7 +33,7 @@
 )
 @click.option(
     "--title",
-    "-c",
+    "-t",
     default=None,
     help="The optional title of the pipeline run.",
 )
```
ruksi: y, probably a copy-paste error

akx:
> Lgtm! Are these automatically added to the docs by the way or does that require another update?

There's no automation on that side of the fence yet :)

PR #371, opened 2026-03-05
Add commands to interact with model catalog

akx:
Agents could be hinted to use `vh` with skills.

A suitably well-written model description is its weight in gold – this makes it nicer to try and ask an agent to, say, "find the latest version of the thingabob price prediction model to answer this query".


ruksi (approved):
adding the model catalog command to Valohai CLI, lgtm

tests and everything, seems proper

PR #373, opened 2026-03-17
Add --reuse option to `vh pipe run`

akx:
Refs #372

Allows replacing selected pipeline nodes with mimic nodes that reference nodes from a previous pipeline run, avoiding re-execution of unchanged steps.

Usage: `vh pipe run MyPipeline --reuse=123:pretrain,mangle`

### DIFF
--- a/valohai_cli/commands/pipeline/run/reuse.py
+++ b/valohai_cli/commands/pipeline/run/reuse.py
@@ -1,0 +1,86 @@
+from __future__ import annotations
+
+import fnmatch
+from typing import Any
+
+import click
+
+from valohai_cli.models.project import Project
+
+
+def parse_reuse_spec(spec: str) -> tuple[str, list[str]]:
+    """
+    Parse a --reuse spec of the form "pipeline-counter:glob1,glob2,...".
+
+    Returns (pipeline_counter, [glob1, glob2, ...]).
+    """
+    if ":" not in spec:
+        raise click.UsageError(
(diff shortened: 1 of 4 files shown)


valohai_cli/commands/pipeline/run/reuse.py
```diff
@@ -0,0 +1,85 @@
...
+        if not matched_any:
+            available = [n["name"] for n in converted_nodes]
+            raise click.UsageError(
+                f"--reuse={spec!r}: no nodes matched the given globs. Available nodes: {sorted(available)}",
+            )
+
+        reused_names = [n["name"] for n in converted_nodes if n.get("type") == "mimic"]
+        click.echo(
+            f"Reusing nodes from pipeline ={counter}: {', '.join(reused_names)}",
+        )
```
ruksi: isn't this echo inside the loop?

wouldn't this cause using multiple reuse specs to incrementally print more and more of mimics?

probably should track them in a separate list and print at the end or something 🤔

ruksi (changes requested):
lgtm otherwise

valohai_cli/commands/pipeline/run/reuse.py
ruksi: oooor track inside the previous matching loop and only print the ones that matched

akx: Fixed!

ruksi (approved):
lgtm!

PR #374, opened 2026-03-30
Become 0.36.0

akx:
Follows up on #370 😄


akx:
Self-merging to get the release rolling.

PR #378, opened 2026-07-09
Add --priority support to `vh task run` and `vh pipe run`

akx:
Shares the existing execution-run priority plumbing: PriorityHackCommand and a new priority_option decorator move to cli_utils so all three run commands declare --priority identically (bare --priority implies 1).

For tasks, priority is passed as a top-level field in the tasks payload. For pipelines, priority is stamped onto every execution/task node's template via apply_node_overrides (which also now hosts the pre-existing environment override).


noorula:
Tested this out of curiosity, works for both pipelines and tasks!

PR #380, opened 2026-07-14
Update and improve CI

akx:
* CI version upgrades
* Hardening with [Zizmor](https://docs.zizmor.sh/)
* Creating GitHub releases when publishing to PyPI


teroyks (approved):
No more forgotten releases

…autogenerated release notes might not be the best, but this could still be a good compromise.

